PR #3, opened 2019-07-02
Add new Button and Label

save-buffer:


CMakeLists.txt
```diff
@@ -9,16 +9,18 @@ add_library(SwiftWin32 SHARED
   Sources/Application/ApplicationDelegate.swift
   Sources/Application/ApplicationMain.swift
   Sources/UI/Color.swift
+  Sources/UI/View.swift
+  Sources/UI/ViewDelegate.swift
   Sources/UI/Window.swift
-  Sources/UI/WindowClass.swift
-  Sources/UI/WindowDelegate.swift)
+  Sources/UI/Button.swift
+  Sources/UI/Label.swift)
```
xiaobai: Nit: sort the sources

Sources/Examples/HelloSwift.swift
```diff
@@ -4,52 +4,63 @@ import SwiftWin32
...
-          SetDCBrushColor(hDC, red.COLORREF)
-          Rectangle(hDC, rctRect.left, rctRect.top, rctRect.right, rctRect.bottom)
-        EndPaint(hWnd, &psPaint)
-
-        return 0
+      return 0
+    }
+      
+    func OnClick(pos: Point) -> LRESULT {
+      print("CLICK WINDOW")
```
xiaobai: Probably don't want this print statement still in here

Sources/Examples/HelloSwift.swift
```diff
@@ -4,52 +4,63 @@ import SwiftWin32
...
+      red = !red
+      if red {
+        self.view?.backgroundColor = #colorLiteral(red: 0.0, green: 0.0, blue: 1.0, alpha: 1.0)
+      } else {
+        self.view?.backgroundColor = #colorLiteral(red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0)
```
xiaobai: Are these conditions supposed to be swapped? It's blue if `red` is true. 😛

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
+
+import WinSDK
+
+typealias WindowProc = @convention(c) (HWND, UINT, WPARAM, LPARAM) -> LRESULT
+let WndProc: WindowProc = { (hWnd, uMsg, wParam, lParam) in
+  let pointer = GetWindowLongPtrW(hWnd, GWLP_USERDATA)
+  if pointer == 0 { print(hWnd); return DefWindowProcW(hWnd, uMsg, wParam, lParam) }
```
compnerd: Debugging left overs?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  let delegate: ViewDelegate =
+    unsafeBitCast(pointer, to: AnyObject.self) as! ViewDelegate
+  switch uMsg {
+  case UINT(WM_CREATE):
+    print("create")
```
compnerd: Debugging left over?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+    return delegate.OnCreate(window: unsafeBitCast(lParam, to: AnyObject.self) as! CREATESTRUCTA)
```
compnerd: Why the forced cast?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  case UINT(WM_DESTROY):
+    return delegate.OnDestroy()
+  case UINT(WM_PAINT):
+    var rctRect: RECT = RECT()
+    GetClientRect(hWnd, &rctRect)
+    var psPaint: PAINTSTRUCT = PAINTSTRUCT()
+    let hDC: HDC = BeginPaint(hWnd, &psPaint)
+    let res = delegate.OnPaint(clientRect: rctRect, hdc: hDC)
+    EndPaint(hWnd, &psPaint)
+    return res
```
compnerd: Why is the default implementation deciding how to render the window?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  case UINT(WM_QUIT):
+    return delegate.OnQuit(exitCode: wParam)
+  case UINT(WM_LBUTTONDOWN):
+    let x = UInt16(lParam & 0xFFFF)
+    let y = UInt16((lParam >> 16) & 0xFFFF)
+    return delegate.OnClick(pos: Point(x: Double(x), y: Double(y)))
```
compnerd: Again, if this this supposed to be the default, why is it defining behaviour?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+
+public class View {
+  public private(set) var subviews = [View]()
+  public private(set) var superview: View? = nil
+  public private(set) var frame: Rect
+  
+  public var hWnd: HWND?
+  internal var wndClassName: String
```
compnerd: This should be named `szClassName` as per the Windows naming convention

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  
+  public var title: String
```
compnerd: This should be named `szTitle` as per the naming convention.  Why does a view have a title?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+  public var style: Int32
```
compnerd: why does this need to be public?  This should be `dwStyle` as per the naming conventions.

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+
+  public var delegate: ViewDelegate {
+    willSet(value) {
+      SetWindowLongPtrW(self.hWnd, GWLP_USERDATA,
+                        unsafeBitCast(value as AnyObject, to: LONG_PTR.self))
+      value.view = self
+    }
+  }
+
+  public init(frame: Rect = Rect(x: Double(CW_USEDEFAULT), y: Double(CW_USEDEFAULT), width: Double(CW_USEDEFAULT), height: Double(CW_USEDEFAULT)),
```
compnerd: Can you wrap this to 80-col please

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+              className: String = "STATIC",
```
compnerd: I think that we should generate a class name and save it, as the name needs to be unique throughout the application.

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+              title: String = "",
```
compnerd: I'm not sure I understand why a view has a title.

Sources/UI/ViewDelegate.swift
```diff
@@ -0,0 +1,73 @@
+
+import WinSDK
+
+public class PaintHelpers {
```
compnerd: I think that this should be `internal`, not `public`.  This shouldn't be part of the API surface.  Can we come up with a better name than `PaintHelpers` please?  Is this supposed to be used outside of `ViewDelegate`?  If not, we could even make this `private`.

Sources/UI/ViewDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+  func OnCreate(window: CREATESTRUCTA) -> LRESULT
+  func OnClose() -> LRESULT
+  func OnDestroy() -> LRESULT
+  func OnPaint(clientRect: RECT, hdc: HDC?) -> LRESULT
+  func OnQuit(exitCode: WPARAM) -> LRESULT
+  func OnClick(pos: Point) -> LRESULT
+}
+
+public extension ViewDelegate {
+
```
compnerd: Why the extra new line here?

Sources/UI/Window.swift
```diff
@@ -1,61 +1,46 @@
-
+
```
compnerd: Bleeding whitespace

CMakeLists.txt
```diff
@@ -9,16 +9,18 @@ add_library(SwiftWin32 SHARED
   Sources/Application/ApplicationDelegate.swift
   Sources/Application/ApplicationMain.swift
   Sources/UI/Color.swift
+  Sources/UI/View.swift
+  Sources/UI/ViewDelegate.swift
```
compnerd: The previous change which moved the delegate around should contain the changes to keep things building.  Each commit should be buildable on their own.

Sources/UI/View.swift
save-buffer: Not necessarily. STATIC is a built-in windows class. If all we want it to do is display a background then this is the built-in class, no need to generate and register a new one.

Sources/Examples/HelloSwift.swift
```diff
@@ -4,52 +4,63 @@ import SwiftWin32
...
-    self.window?.delegate = SwiftWindowDelegate()
-    self.window?.show()
+    self.btn?.backgroundColor = #colorLiteral(red: 0.0, green: 0.0, blue: 1.0, alpha: 1.0)
+    
+    self.btn?.text = "BUTTON HI"
+    self.label?.text = "LABEL HI"
+    self.label?.textColor = #colorLiteral(red: 0.0, green: 1.0, blue: 0.0, alpha: 1.0)
+    self.label?.borderThickness = 10
 
+    self.window?.addSubview(self.btn ?? Button(title: ""))
```
adierking: I think it would be more concise to unwrap the button and label here instead of using `??`. (Perhaps we could also store the unwrapped values to variables instead of using `self.btn?` everywhere?)
```suggestion
    self.window?.addSubview(self.btn!)
```

Sources/UI/View.swift
save-buffer: I don't really see what you mean - every single OnPaint method that I can think of will be doing this BeginPaint/EndPaint pattern, so might as well factor it out.

Sources/UI/View.swift
save-buffer: It's not defining any behavior - it's just unpacking arguments and making it nicer for the user to define their own delegates

Sources/UI/Button.swift
```diff
@@ -0,0 +1,30 @@
...
+}
+
+public class Button: View {
+  public var textColor: Color = #colorLiteral(red: 0.0, green: 0.0, blue: 0.0, alpha: 1.0)
+  public var text: String = ""
+  public init(frame: Rect = Rect(x: 100, y: 100, width: 100, height: 100), title: String) {
+    super.init(frame: frame,
+               className: "BUTTON",
+               title: title,
+               style: WS_TABSTOP | WS_VISIBLE | BS_DEFPUSHBUTTON | BS_TEXT | BS_VCENTER)
```
adierking: I don't remember exactly what the differences between `BS_DEFPUSHBUTTON` and `BS_PUSHBUTTON` are, but I think we should use `BS_PUSHBUTTON` instead.

Sources/UI/View.swift
save-buffer: Window uses the Title as a class name. I agree that that may be problematic if you want multiple windows with the same title but with different styles (in that case, we should generate a class name in Window).

Sources/UI/Button.swift
```diff
@@ -0,0 +1,30 @@
...
+        PaintHelpers.DrawText(clientRect: clientRect, hdc: hdc, text: btn.text, color: fg)
+    }
+    return 0
+  }
+}
+
+public class Button: View {
+  public var textColor: Color = #colorLiteral(red: 0.0, green: 0.0, blue: 0.0, alpha: 1.0)
+  public var text: String = ""
+  public init(frame: Rect = Rect(x: 100, y: 100, width: 100, height: 100), title: String) {
```
adierking: What is the purpose of the `title` parameter here?

Sources/UI/View.swift
adierking: Yeah, I think that the `BeginPaint()` and `EndPaint()` calls should be left up to the delegate. When we invoke delegate methods, we should only be responsible for marshalling the parameters and return value.

Sources/UI/View.swift
adierking: What benefits do we get by using `STATIC` instead of a custom class?

Sources/UI/View.swift
```diff
@@ -0,0 +1,102 @@
...
+              style: Int32 = 0) {
+    self.frame = frame
+    self.wndClassName = className
+    self.title = title
+    self.style = style
+    self.hWnd = CreateWindowExW(0, self.wndClassName.LPCWSTR, self.title.LPCWSTR,
```
adierking: Should we expose `dwExStyle` instead of forcing it to 0?

Sources/UI/View.swift
save-buffer: Not really sure, I think I copied it from elsewhere in the code. I was trying to cast the lParam to a pointer to a CREATESTRUCTA. Is there a better way? (or should I change everything back to take raw lparam and wparam?)

Sources/UI/View.swift
save-buffer: Because CreateWindowExW needs a title

Sources/UI/ViewDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+import WinSDK
+
+public class PaintHelpers {
+  public static func DrawBackground(clientRect: RECT, hdc: HDC?, color: Color) {
```
adierking: I'm not sure that this needs to have `Background` in its name since it basically fills a rectangle. Maybe `FillRectangle`?

Sources/UI/ViewDelegate.swift
```diff
@@ -0,0 +1,73 @@
...
+    var local = window
+    return DefWindowProcW(self.view?.hWnd, UINT(WM_CREATE), 0, LPARAM(Int(bitPattern: UnsafeRawPointer(&local))))
+  }
+
+  func OnClose() -> LRESULT {
+    return DefWindowProcW(self.view?.hWnd, UINT(WM_CLOSE), 0, 0)
+  }
+  
+  func OnDestroy() -> LRESULT {
+    PostQuitMessage(0)
```
adierking: I don't think that we should use `PostQuitMessage()` by default. It should only be used with the main window.

Sources/UI/ViewDelegate.swift
save-buffer: Could you give me some qualities that you'd like the new name to exhibit? I feel that `PaintHelpers` adequately describes what this class contains.
I'll change it to internal - it is used in default implementations for Label and Button.

Sources/UI/Window.swift
```diff
@@ -1,61 +1,46 @@
...
-    SetWindowLongPtrW(hWnd, GWLP_WNDPROC,
-                      unsafeBitCast(WndProc, to: LONG_PTR.self))
-  }
-
-  public func show() {
-    ShowWindow(self.hWnd, SW_SHOW)
+public class Window: View {
+  var windowClass: WindowClass
+  public init(title: String) {
+    self.windowClass = WindowClass(hInst: GetModuleHandleW(nil), name: title)
```
adierking: I don't think we should use the window title as the class name. There's a risk of accidentally using an already-registered class name, and class names cannot be longer than 256 characters. As Saleem mentioned, we should figure out a better way to generate class names.

Sources/UI/View.swift
save-buffer: Well, I think the more appropriate question is what benefits do we get by generating a custom class for every view? I guess I don't really see a reason to...

Sources/UI/View.swift
compnerd: I think that we could do an extension to the `ViewDelegate` but leave the `ViewDelegate` as the stock WndProc.  WDYT?

Sources/UI/View.swift
compnerd: Each class is something which needs to be unique.  All labels would be of a single class so you cannot actually enumerate them individually.  The window delegate should be the same for them all, which again seems wrong.  The question is not, what benefit do you get by creating a custom class, its why would you not do that?

Sources/UI/View.swift
compnerd: I think that we should pass `nil` to that, and use `SetWindowTextW` for the title.

Sources/UI/ViewDelegate.swift
compnerd: I generally dislike "Helper", "Utilities", etc.  They are superfluous and add nothing.  What type of helpers are these?  I would like something pithy and precise.

Sources/UI/View.swift
```diff
@@ -7,8 +7,8 @@ let WndProc: WindowProc = { (hWnd, uMsg, wParam, lParam) in
   let pointer = GetWindowLongPtrW(hWnd, GWLP_USERDATA)
   if pointer == 0 { return DefWindowProcW(hWnd, uMsg, wParam, lParam) }
 
-  let delegate: WindowDelegate =
-      unsafeBitCast(pointer, to: AnyObject.self) as! WindowDelegate
+  let delegate: ViewDelegate =
+      unsafeBitCast(pointer, to: AnyObject.self) as! ViewDelegate
```
compnerd: Bleeding whitespace

Sources/Examples/HelloSwift.swift
```diff
@@ -41,7 +41,7 @@ class SwiftApplicationDelegate: ApplicationDelegate {
       }
     }
 
-    self.window = View(class: SwiftWindowClass, title: "Swift Window")
+    self.window = Window(wndClass: SwiftWindowClass, szTitle: "Swift Window")
```
compnerd: Why the change in parameter names?  I think that `Wndow(class:title:)` is a reasonable signature.  Is there a motivation for renaming these?

Sources/UI/View.swift
```diff
@@ -36,24 +40,54 @@ public struct View {
...
-  public init(`class`: WindowClass, title: String, hInstance: HINSTANCE? = nil,
-              hWndParent: HWND? = nil) {
-    self.class = `class`
-    title.withCString(encodedAs: UTF16.self) {
-      self.hWnd = CreateWindowExW(0, self.class.name, $0,
-                                  WS_OVERLAPPEDWINDOW,
-                                  CW_USEDEFAULT, CW_USEDEFAULT,
-                                  CW_USEDEFAULT, CW_USEDEFAULT,
-                                  hWndParent, nil, hInstance, nil)
+  public var backgroundColor: Color = #colorLiteral(red: 1.0, green: 0.0, blue: 0.0, alpha: 1.0) {
```
compnerd: I think that we can default to black or white, or even better, the themes default background color.  The red was mostly for testing purposes.

Sources/UI/View.swift
```diff
@@ -36,24 +40,54 @@ public struct View {
...
+  internal var hWnd: HWND?
+  internal var wndClass: WindowClass
+  internal var dwStyle: Int32
+  
+  public init(frame: Rect = Rect(x:      Double(CW_USEDEFAULT),
+                                 y:      Double(CW_USEDEFAULT),
+                                 width:  Double(CW_USEDEFAULT),
+                                 height: Double(CW_USEDEFAULT)),              
+              wndClass: WindowClass, 
+              dwStyle: Int32 = 0) {
```
compnerd: I think that I prefer the signature `View(frame:class:style:)`

Sources/UI/View.swift
```diff
@@ -36,24 +40,54 @@ public struct View {
...
+  }
 
+  internal var hWnd: HWND?
+  internal var wndClass: WindowClass
+  internal var dwStyle: Int32
+  
+  public init(frame: Rect = Rect(x:      Double(CW_USEDEFAULT),
+                                 y:      Double(CW_USEDEFAULT),
+                                 width:  Double(CW_USEDEFAULT),
+                                 height: Double(CW_USEDEFAULT)),
```
compnerd: I think that it might be nice to extract this into an internal extension and use that, something like this:

```
internal extension Rect {
  static let default: Rect = Rect(x: Double(CW_USEDEFAULT),
                                  y: Double(CW_USEDEFAULT),
                                  width: Double(CW_USEDEFAULT),
                                  height: Double(CW_USEDEFAULT))
}
```

```
  pubic init(frame: Rect = .default, class: WindowClass, style: Int32 = 0)
```

Sources/UI/View.swift
```diff
@@ -36,24 +40,54 @@ public struct View {
...
+  }
+
+  public func addSubviews(_ views: [View]) {
+    for v in views {
+      addSubview(v)
+    }
+  }
+
+  public func invalidate() {
+      RedrawWindow(self.hWnd, nil, nil, UINT(RDW_INVALIDATE))
```
compnerd: 2 space indent place

Sources/UI/Window.swift
```diff
@@ -0,0 +1,28 @@
+	
+import WinSDK
+
+public class Window: View {
+  public var szTitle: String {
```
compnerd: Lets just make this `title`.

Sources/UI/Window.swift
```diff
@@ -0,0 +1,28 @@
...
+      didSet(value) {
+          SetWindowTextW(self.hWnd, value.LPCWSTR)
```
compnerd: 2 space indent please

Sources/UI/Window.swift
```diff
@@ -0,0 +1,28 @@
...
+      }
+  }
+  
+  public init(wndClass: WindowClass? = nil,
```
compnerd: I think that the signature should be `Window(class:title:style:)`.  For Window, I don't think that we want to make title optional.

Sources/UI/Window.swift
```diff
@@ -0,0 +1,28 @@
...
+              dwStyle: UInt32 = WS_OVERLAPPEDWINDOW,
+              szTitle: String = "") {
+    struct Counter {
+      static var windowsCreated: UInt64 = 0
```
compnerd: `value` seems a more pithy name

Sources/UI/Window.swift
```diff
@@ -0,0 +1,28 @@
...
+    self.szTitle = szTitle
+    let wnd = wndClass ?? WindowClass(hInst: GetModuleHandleW(nil),
+                                      name: "w\(Counter.windowsCreated)")
+    super.init(wndClass: wnd, dwStyle: Int32(WS_OVERLAPPEDWINDOW))
+    
+    Counter.windowsCreated += 1
+  }
+
+  deinit {
+      _ = self.wndClass.unregister()
```
compnerd: What happens if the class was not generated and more than one window was created with this class?

Sources/UI/Button.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public class Button: View {
+  public var textColor: Color = #colorLiteral(red: 0.0, green: 0.0, blue: 0.0, alpha: 1.0)
+  public var text: String = "" {
+    didSet(value) {
+      self.invalidate()
+    }
+  }
+  
+  public private(set) static var defaultWindowClass: WindowClass =
```
compnerd: Hmm, why do we want this to be public?

Sources/UI/Button.swift
```diff
@@ -0,0 +1,26 @@
...
+    WindowClass(hInst: GetModuleHandleW(nil),
+                name: "UI.Button")
+  
+  public init(frame: Rect = Rect(x: 100,
+                                 y: 100,
+                                 width: 100,
+                                 height: 100),
```
compnerd: Please bin-pack the arguments.

Sources/UI/Button.swift
```diff
@@ -0,0 +1,26 @@
...
+              dwStyle: Int32 = WS_TABSTOP | WS_VISIBLE | BS_PUSHBUTTON | BS_TEXT | BS_VCENTER,
+              wndClass: WindowClass = defaultWindowClass) {
```
compnerd: I think that the signature should be `Button(frame:style:class:)`.

Sources/UI/Button.swift
```diff
@@ -0,0 +1,26 @@
...
+    super.init(frame: frame,
+               wndClass: wndClass,
+               dwStyle: dwStyle)
```
compnerd: Please bin-pack the arguments

Sources/UI/Label.swift
```diff
@@ -0,0 +1,37 @@
...
+      self.invalidate()
+    }
+  }
+  public var text: String = "" {
+    didSet(value) {
+      self.invalidate()
+    }
+  }
+  
+  public private(set) static var defaultWindowClass: WindowClass =
```
compnerd: Similar to the button class

Sources/UI/WindowClass.swift
```diff
@@ -6,6 +6,8 @@ public class WindowClass {
   internal var `class`: WNDCLASSEXW
   internal var hInstance: HINSTANCE
 
+  public private(set) var registered: Bool = false
```
compnerd: Hmm, do you expect people wanting to check if the class is registered?  Or can this just be private?

Sources/Examples/HelloSwift.swift
```diff
@@ -6,58 +6,54 @@ class SwiftApplicationDelegate: ApplicationDelegate {
...
-        let hDC: HDC = BeginPaint(hWnd, &psPaint)
-          SelectObject(hDC, GetStockObject(NULL_PEN))
-          SelectObject(hDC, GetStockObject(DC_BRUSH))
-          SetDCBrushColor(hDC, red.COLORREF)
-          Rectangle(hDC, rctRect.left, rctRect.top, rctRect.right, rctRect.bottom)
-        EndPaint(hWnd, &psPaint)
-
-        return 0
-      }
+  class MyWndDelegate: WindowDelegate {
```
compnerd: I think `HelloSwiftWindowDelegate` is nicer.

Sources/UI/Button.swift
```diff
@@ -1,6 +1,20 @@
 
 import WinSDK
 
+open class ButtonDelegate: DefaultDelegate {
+  open override func OnPaint(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT {
+      if let view = view {
```
compnerd: I think you can do:

```
guard let view = view else { return 0 }
```

This will let you unindent the body of the code

Sources/UI/Label.swift
```diff
@@ -1,6 +1,23 @@
 
 import WinSDK
 
+open class LabelDelegate: DefaultDelegate {
+  open override func OnPaint(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT {
+    if let view = self.view {
```
compnerd: Similar

Sources/UI/ViewDelegate.swift
```diff
@@ -2,10 +2,13 @@
 import WinSDK
 
 public protocol ViewDelegate: class {
+  var view: View? { get set }
+  
   func OnCreate(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT
   func OnDestroy(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT
   func OnPaint(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT
   func OnQuit(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT
+  func OnLButtonDown(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT
```
compnerd: I think we can spell out `Left`

Sources/UI/ViewDelegate.swift
```diff
@@ -2,10 +2,13 @@
 import WinSDK
 
 public protocol ViewDelegate: class {
+  var view: View? { get set }
```
compnerd: I really wish that we didn't have to have this member in the delegate.  Is it not possible to retrieve the view from the `hWnd`?

Sources/UI/ViewDelegate.swift
```diff
@@ -24,5 +27,39 @@ public extension ViewDelegate {
   func OnQuit(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT {
     return DefWindowProcW(hWnd, UINT(WM_QUIT), wParam, lParam)
   }
+  
+  func OnLButtonDown(_ hWnd: HWND, _ wParam: WPARAM, _ lParam: LPARAM) -> LRESULT {
+    return DefWindowProcW(hWnd, UINT(WM_LBUTTONDOWN), wParam, lParam)
+  }
 }
 
+open class DefaultDelegate: ViewDelegate {
```
compnerd: I don't understand why we need this.  The extension to the `ViewDelegate` will provide the default implementation, and we can just call the default window proc if no delegate is set.

Sources/Examples/HelloSwift.swift
save-buffer: I didn't like the way the backticks looked :P
I'll change it back

Sources/UI/ViewDelegate.swift
save-buffer: To provide `var view: View?`, which you can't do in a protocol.

Sources/UI/ViewDelegate.swift
save-buffer: I don't know of a way, right now it's only set up so that you can retrieve the delegate from the hWnd, but not the actual view.

Sources/UI/Button.swift
save-buffer: So that we can use it as a default argument

Sources/UI/View.swift
save-buffer: Sounds good, doesn't work. If we want to use as a default argument for the public constructor, we have to make it public.

Sources/UI/ViewDelegate.swift
compnerd: Oh, we can already do that!  `GetWindowLongPtr`

compnerd:
Thanks, this looks pretty good.  Could you please rework the last change into the changes into the original commits that introduced the change.  At that point, this is ready to be merged :-).

Sources/UI/View.swift
```diff
@@ -0,0 +1,97 @@
...
+    self.class = `class`
+    _ = self.class.register()
+    self.style = style
+    self.hWnd = CreateWindowExW(0, self.class.name, nil,
+                                DWORD(self.style),
+                                Int32(self.frame.x), Int32(self.frame.y),
+                                Int32(self.frame.width), Int32(self.frame.height),
+                                nil, nil, GetModuleHandleW(nil), nil)
+    self.delegate = DefaultDelegate()
+    SetWindowLongPtrW(self.hWnd, GWLP_USERDATA,
```
adierking: I'm not sure that this call is necessary. Shouldn't setting `self.delegate` change `GWLP_USERDATA` for us?

Sources/UI/View.swift
```diff
@@ -0,0 +1,97 @@
...
+                      unsafeBitCast(self.delegate as AnyObject, to: LONG_PTR.self))
+    SetWindowLongPtrW(hWnd, GWLP_WNDPROC,
```
adierking: Can we make this come from the class instead of forcing it? Otherwise we lose a pretty big advantage to being able to specify a custom class.

Sources/UI/View.swift
save-buffer: willSet and didSet don't run from init():
> The willSet and didSet observers of superclass properties are called when a property is set in a subclass initializer, after the superclass initializer has been called. They are not called while a class is setting its own properties, before the superclass initializer has been called.

Sources/UI/View.swift
save-buffer: I think this functionality can be added using Swift classes instead. For example, if you wanted a bunch of buttons with the same delegate, just subclass Button and in its constructor change it to what you want.

Sources/UI/View.swift
adierking: ahh, gotcha - sorry, I don't know Swift that well :(

PR #6, opened 2019-10-17
Window changes

farzonl:
## CMakeLists.txt:
- add COBJMACROS forces the C API instead of the C++ api

## View.swift:
- add AdjustWindowRect for window creation so that panel is the size of
rect excluding the border size
- Make hWind a public getter private setter

## WindowClass.swift:
- expose style and hbrBackground in the constructor

## HelloSwift.swift:
- Add a sub window to test both Window class constructor changes and View AdjustWindowRect change
- Make top level window redraw on horizontal and vertical size chanes
- set sub window to a different color
- add label to the sub window

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ target_sources(SwiftWin32 PRIVATE
 target_sources(SwiftWin32 PRIVATE
   Sources/Support/Rect+UIExtensions.swift
   Sources/Support/String.swift)
+add_definitions(-DCOBJMACROS)
 target_link_libraries(SwiftWin32 PUBLIC User32 Gdi32 ComCtl32)
 set_target_properties(SwiftWin32 PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY}
--- a/Sources/Examples/HelloSwift.swift
+++ b/Sources/Examples/HelloSwift.swift
@@ -30,6 +30,14 @@
 import WinSDK
 import SwiftWin32
 
+class SubWindowEventHandler: WindowDelegate {
+  func OnDestroy(_ hWnd: HWND?, _ wParam: WPARAM, _ lParam: LPARAM)
+      -> LRESULT {
+    PostQuitMessage(0)
+    return 0
+  }
+}
+
 class EventHandler: WindowDelegate {
   func OnDestroy(_ hWnd: HWND?, _ wParam: WPARAM, _ lParam: LPARAM)
       -> LRESULT {
@@ -46,21 +54,34 @@ class EventHandler: WindowDelegate {
 }
 
 class SwiftApplicationDelegate: ApplicationDelegate {
+  public static let `parent`: WindowClass =
+      WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Top.Window",
+                  style : DWORD(CS_HREDRAW | CS_VREDRAW), hbrBackground : DWORD(COLOR_WINDOW) )
   lazy var window: Window =
-      Window(frame: .default, title: "Swift/Win32 Window")
+      Window(frame: .default, class: SwiftApplicationDelegate.parent, title: "Swift/Win32 Window")
+  lazy var rect = RECT();
+	
+  lazy var subwindow: Window =
+       Window(frame:Rect(x: 0, y: 0, width: Double(rect.right / 2), height: Double(rect.bottom/2)), title: "sub window")
   lazy var button: Button =
-      Button(frame: Rect(x: 64, y: 0, width: 64, height: 32), title: "Press Me!")
+      Button(frame: Rect(x: 864, y: 0, width: 64, height: 32), title: "Press Me!")
   lazy var label: Label =
       Label(frame: Rect(x: 0, y: 0, width: 64, height: 8), title: "Read Me:")
   lazy var progress: ProgressView =
       ProgressView(frame: Rect(x: 0, y: 48, width: 256, height: 32))
   lazy var delegate = EventHandler()
+  lazy var subWindowdelegate = SubWindowEventHandler()
 
   func application(_: Application,
                    didFinishLaunchingWithOptions options: [Application.LaunchOptionsKey:Any]?) -> Bool {
-    window.addSubview(self.label)
+    GetClientRect(window.hWnd, &rect)
+    let brush = CreateSolidBrush(DWORD(255))
+    SetClassLongPtrW(subwindow.hWnd, GCLP_HBRBACKGROUND, unsafeBitCast(brush, to: Int64.self))
+    subwindow.addSubview(self.label)
+    window.addSubview(self.subwindow)
     window.addSubview(self.button)
-    window.addSubview(self.progress)
+    subwindow.addSubview(self.progress)
+    subwindow.delegate = subWindowdelegate
     window.delegate = delegate
     self.progress.setProgress(0.5, animated: false)
     return true
--- a/Sources/UI/View.swift
+++ b/Sources/UI/View.swift
@@ -30,7 +30,7 @@
 import WinSDK
 
 public class View {
-  internal var hWnd: HWND
+  public private(set) var hWnd: HWND
   internal var `class`: WindowClass
   internal var style: DWORD
 
@@ -45,6 +45,18 @@ public class View {
     self.style = style
 
     self.frame = frame
+    if(self.frame.width != Double(CW_USEDEFAULT)) {
+      var r = RECT(left:   Int32(self.frame.x), 
+                   top:    Int32(self.frame.y),
+                   right:  Int32(self.frame.width),
+                   bottom: Int32(self.frame.height))
+              
+      AdjustWindowRect(&r, self.style, false)
+      self.frame.x = Double(r.left)
+      self.frame.y = Double(r.top)
+      self.frame.width = Double(r.right - r.left)
+      self.frame.height = Double(r.bottom-r.top)
+    }
     self.hWnd =
         CreateWindowExW(0, self.class.name, "".LPCWSTR, self.style,
                         Int32(self.frame.x), Int32(self.frame.y),
--- a/Sources/UI/WindowClass.swift
+++ b/Sources/UI/WindowClass.swift
@@ -34,25 +34,25 @@ public typealias WindowProc =
 
 public class WindowClass {
   internal var name: [WCHAR]
-
   internal var hInstance: HINSTANCE?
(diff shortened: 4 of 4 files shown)


Sources/Examples/HelloSwift.swift
```diff
@@ -46,21 +54,34 @@ class EventHandler: WindowDelegate {
 }
 
 class SwiftApplicationDelegate: ApplicationDelegate {
+  public static let `parent`: WindowClass =
+      WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Top.Window",
+                  style : DWORD(CS_HREDRAW | CS_VREDRAW), hbrBackground : DWORD(COLOR_WINDOW) )
   lazy var window: Window =
-      Window(frame: .default, title: "Swift/Win32 Window")
+      Window(frame: .default, class: SwiftApplicationDelegate.parent, title: "Swift/Win32 Window")
```
compnerd: Why the custom class?  Is it for the background?  If so, why not set that for the actual Swift Window class?

Sources/Examples/HelloSwift.swift
```diff
@@ -46,21 +54,34 @@ class EventHandler: WindowDelegate {
...
+  lazy var rect = RECT();
+	
+  lazy var subwindow: Window =
+       Window(frame:Rect(x: 0, y: 0, width: Double(rect.right / 2), height: Double(rect.bottom/2)), title: "sub window")
   lazy var button: Button =
-      Button(frame: Rect(x: 64, y: 0, width: 64, height: 32), title: "Press Me!")
+      Button(frame: Rect(x: 864, y: 0, width: 64, height: 32), title: "Press Me!")
```
compnerd: Why the huge shift?

Sources/Examples/HelloSwift.swift
```diff
@@ -46,21 +54,34 @@ class EventHandler: WindowDelegate {
...
   lazy var label: Label =
       Label(frame: Rect(x: 0, y: 0, width: 64, height: 8), title: "Read Me:")
   lazy var progress: ProgressView =
       ProgressView(frame: Rect(x: 0, y: 48, width: 256, height: 32))
   lazy var delegate = EventHandler()
+  lazy var subWindowdelegate = SubWindowEventHandler()
```
compnerd: What is the reasoning for the sub-window?

Sources/UI/View.swift
```diff
@@ -30,7 +30,7 @@
 import WinSDK
 
 public class View {
-  internal var hWnd: HWND
+  public private(set) var hWnd: HWND
```
compnerd: Why make this public?

Sources/UI/View.swift
```diff
@@ -45,6 +45,18 @@ public class View {
     self.style = style
 
     self.frame = frame
+    if(self.frame.width != Double(CW_USEDEFAULT)) {
+      var r = RECT(left:   Int32(self.frame.x),
```
compnerd: Is this a common operation?

Sources/UI/View.swift
```diff
@@ -45,6 +45,18 @@ public class View {
     self.style = style
 
     self.frame = frame
+    if(self.frame.width != Double(CW_USEDEFAULT)) {
+      var r = RECT(left:   Int32(self.frame.x), 
+                   top:    Int32(self.frame.y),
+                   right:  Int32(self.frame.width),
+                   bottom: Int32(self.frame.height))
+              
+      AdjustWindowRect(&r, self.style, false)
```
compnerd: Is this to clip the rect?

Sources/UI/WindowClass.swift
```diff
@@ -34,25 +34,25 @@ public typealias WindowProc =
...
   internal var name: [WCHAR]
-
   internal var hInstance: HINSTANCE?
   internal var value: WNDCLASSEXW?
   internal var atom: ATOM?
 
   public init(hInst hInstance: HINSTANCE, name: String,
-              WindowProc lpfnWindowProc: WindowProc? = DefWindowProcW) {
+              WindowProc lpfnWindowProc: WindowProc? = DefWindowProcW,
+              style : UInt32 = 0, hbrBackground : UInt32 = 0 ) {
```
compnerd: No space before the `:`.

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ target_sources(SwiftWin32 PRIVATE
 target_sources(SwiftWin32 PRIVATE
   Sources/Support/Rect+UIExtensions.swift
   Sources/Support/String.swift)
+add_definitions(-DCOBJMACROS)
```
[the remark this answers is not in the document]
farzonl: sure.

Sources/Examples/HelloSwift.swift
farzonl: this is a typo should be Double(rect.right / 2) + a few pixels to be outside of sub-window.

Sources/UI/View.swift
farzonl: So for windows this is a common operation as often when you ask windows for a 800:600 window for example you don't expect the boarder to be factored into that calculation. passing the size you want to AdjustWindowRect allows us to account for the boarder size. It should not have an effect on buttons or labels.

Sources/Examples/HelloSwift.swift
farzonl: So the custom class was done more than to change a background color.  This was something I found weird about the current implementation. It is my feeling that you should not have a static window class defined in class Window. If you have a static window class then you can't ever create sub windows b\c any change made to the static Windowclass will cascade. For example if I had not done this  then when I made the subwindow red it would have made the parent red as well.  Instead each class Window should have their own Windowclass instance.

Sources/Examples/HelloSwift.swift
farzonl: So  applications like excel are multiple windows deep. I figured the example should test this behavior. Further it gave me a means of testing both a  window with a CW_USEDEFAULT window size and one with a custom window size. So the addition was also to exercise more control flow logic.

Sources/UI/View.swift
farzonl: Hwnds are typically a publicly accessible property. They get register with the os so hiding them doesn't do you much good. a tool like spy++ or reflection with c# could expose the property.

# examples of public apis
- https://docs.microsoft.com/en-us/office/vba/api/word.window.hwnd
- https://docs.microsoft.com/en-us/office/vba/api/access.form.hwnd
- https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.handle?view=netframework-4.8

Another reason is if you want to mix win32 code and  swift-win32 many of the apis in win32 need you to pass in an hwnd. Exposing a public getter would  at least mean you don't have to write as much interop code to use these win32 api functions.

Sources/UI/View.swift
farzonl: should actually grow the rect.

Sources/UI/WindowClass.swift
farzonl: ah right, ill fix that.

compnerd:
Personally, I would recommend pulling this apart into a series of patches, so we can get some of the items merged more quickly.

Sources/Examples/HelloSwift.swift
compnerd: Thats fair.  I wanted to simplify the construction of a single window application.  Is there a way to create a top level window without having to create the custom class and let the system synthesize one as appropriate?

Sources/Examples/HelloSwift.swift
compnerd: Ah, that makes much more sense.  I think personally, what I would prefer is that we instead introduce a new type, `Insets` which would allow us to have an arbitrary sized inset for any view.

Sources/Examples/HelloSwift.swift
compnerd: Oh, I see.  Can you please add a comment explaining the reasoning for this.  I don't really have an issue with making this code more complicated to represent real-world scenarios.  I think that we may want to introduce a second demo program though so that we can represent both the most trivial usage as well as the real world usage - "HelloSwift" would be the equivalent of a "Hello World" program, and then a real world application could be a "SwiftDemo" or something.

I don't know how how much time you have, or if you would even be up for writing something like this, but I would *love* to have a complex demo with multi-level windows, possibly with a custom Widget, maybe with a DX layer, rendered with a 2000-styled menu bar and metro widgets!

Sources/UI/View.swift
compnerd: Yeah, Im kinda torn on this is the thing.  Im happy with making the `hWnd` a SPI for now, something like:

```
public extension View {
  var _handle: HWND { return hWnd }
}
```

Yes, the hWnd is easy to get - spy++ is part of the DDK tooling, so its generally available and will give you all the details of the Window.  But, at the same time, I would like to make it so that users of Swift/Win32 don't really need to fully understand all the details of the Windowing system in Windows, and can quickly get productive in writing applications.  

@plotfi actually recently experienced this and was pleasantly surprised at how easy it was to write a UI application with this.

Sources/UI/View.swift
compnerd: If this is common, perhaps it is better to do something like introduce a new file (`RECT+UIExtensions.swift` or something?) with custom extensions for this.

```
internal extension RECT {
  convenience init(from rect: Rect) {
  }
}
```

Or perhaps even a computed property in `Rect` to get a `RECT` from it?  (That could easily fit into `Rect+UIExtensions.swift`)

CMakeLists.txt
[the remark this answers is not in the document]
farzonl: https://github.com/compnerd/swift-win32/pull/8

Sources/UI/WindowClass.swift
farzonl: https://github.com/compnerd/swift-win32/pull/7

farzonl:
> Personally, I would recommend pulling this apart into a series of patches, so we can get some of the items merged more quickly.

created two PRs for CMakeLists.txt and WindowClass.swift. will put out a change for the rect extension later this week.  I'll hold off on any hwnd accessor changes for now. I can make a separate example that does sub-windowing, not sure the scope of work for Insets so ill hold off on that as well.

Sources/Examples/HelloSwift.swift
farzonl: I'll think about it. At the moment I was thinking maybe there should be some factory method that takes in an enum  and returns back a predefined WindowClass  construction. Then there can be two window constructors one that takes in a class and one that takes in the enum to generate the WindowClass.

PR #9, opened 2019-10-24
Add Rect extension functions and window panel size adjustments.

farzonl:
## Rect+UIExtensions.swift:
- computed property RECT getter
- RECT setter as asignment operator

## View.swift:
- using Rect extentions call AdjustWindowRect to set window size so that panel is the size of rect excluding the border size.

### DIFF
--- a/Sources/Support/Rect+UIExtensions.swift
+++ b/Sources/Support/Rect+UIExtensions.swift
@@ -34,3 +34,25 @@ internal extension Rect {
       Rect(x: Double(CW_USEDEFAULT), y: Double(CW_USEDEFAULT),
            width: Double(CW_USEDEFAULT), height: Double(CW_USEDEFAULT))
 }
+
+public extension Rect {
+  init(from: RECT) {
+    self.x = Double(from.left)
+    self.y = Double(from.top)
+    self.width  = Double(from.right - from.left)
+    self.height = Double(from.bottom - from.top)
+  }
+
+  var isAnyPointDefault: Bool {
+    return self.x == Double(CW_USEDEFAULT) || self.y == Double(CW_USEDEFAULT) ||
+           self.width == Double(CW_USEDEFAULT) || 
+           self.height == Double(CW_USEDEFAULT)
+  }
+}
+
+internal extension RECT {
+  init(from: Rect) {
+    self.init(left: Int32(from.x), top: Int32(from.y), right: Int32(from.width),
+              bottom: Int32(from.height))
+  }
+}
--- a/Sources/UI/View.swift
+++ b/Sources/UI/View.swift
@@ -45,6 +45,11 @@ public class View {
     self.style = style
 
     self.frame = frame
+    if !self.frame.isAnyPointDefault {
+      var r = RECT(from: self.frame);
+      AdjustWindowRect(&r, self.style, false)
+      self.frame = Rect(from: r)
+    }
     self.hWnd =
         CreateWindowExW(0, self.class.name, "".LPCWSTR, self.style,
                         Int32(self.frame.x), Int32(self.frame.y),


Sources/Support/Rect+UIExtensions.swift
```diff
@@ -29,8 +29,22 @@
...
            width: Double(CW_USEDEFAULT), height: Double(CW_USEDEFAULT))
+  
+  var getRECT: RECT {
+      return RECT(left:   Int32(self.x), 
+                  top:    Int32(self.y),
+                  right:  Int32(self.width),
+                  bottom: Int32(self.height))
+  }
+
+  static func &= (l: inout Rect, r: RECT) {
```
farzonl: from what I have read, swift does not seem to let you override the assignment operator so I'm using &= instead. hope this is ok.

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -29,8 +29,22 @@
 
 import WinSDK
 
-internal extension Rect {
-  public static let `default`: Rect =
+public extension Rect {
```
compnerd: I think that you should do a second extension, one that is `public` rather than `internal`, so that we do not expose `default` to users of the library.  I don't think that we should encourage the use of that, preferring `.zero` instead

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -29,8 +29,22 @@
...
+  static let `default`: Rect =
       Rect(x: Double(CW_USEDEFAULT), y: Double(CW_USEDEFAULT),
            width: Double(CW_USEDEFAULT), height: Double(CW_USEDEFAULT))
+  
+  var getRECT: RECT {
```
compnerd: Make the computed property `RECT` instead.  Alternatively, do an extension on `RECT` which does a convenience named initializer.

Sources/Support/Rect+UIExtensions.swift
compnerd: Why the need for the assignment?  Why not make it explicit with a convenience initializer?

Sources/UI/View.swift
```diff
@@ -45,6 +45,11 @@ public class View {
     self.style = style
 
     self.frame = frame
+    if(self.frame.width != Double(CW_USEDEFAULT)) {
```
compnerd: Space after the `if`.  I'm not sure I understand the condition here.

farzonl: CW_USEDEFAULT is a special  parameter that doesn't corresponds to screen coordinates. without this conditional I wouldn't be able to do the width and height calculations i.e r.right - r.left and r.bottom-r.top

Sources/Support/Rect+UIExtensions.swift
farzonl: i figured an initializer would allocate memory, but assignment would not. I can change to convenience.

Sources/Support/Rect+UIExtensions.swift
farzonl: So if the extension is internal and default is public isn't it already exposed?.  In any case I can make this change.

farzonl: setting it back to the way it was I now see the following warning: "warning: 'public' modifier conflicts with extension's default access of 'internal'
  public static let `default`: Rect ="

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -34,3 +34,21 @@ internal extension Rect {
...
+  init(rect : RECT) {
+    self.x = Double(rect.left)
+    self.y = Double(rect.top)
+    self.width  = Double(rect.right - rect.left)
+    self.height = Double(rect.bottom - rect.top)
+  }
+}
+
+public extension RECT {
+  init( rect : Rect) {
```
farzonl: was not able to make this a convenience init b\c RECT is a constructor.

Sources/Support/Rect+UIExtensions.swift
compnerd: Yeah, the convenience init seems better.  The allocations shouldn't matter - in optimized mode, it should be possible to see that move semantics should apply here.

Sources/Support/Rect+UIExtensions.swift
compnerd: Correct. You would need a *second* extension in the same file.

Sources/UI/View.swift
compnerd: But, why are you just checking if the width is "CW_USEDEFAULT" rather than all of them?

Sources/Support/Rect+UIExtensions.swift
compnerd: Ah, okay, a regular initializer is fine.  Please make the init actually have the signature `init(from:)`.  This is effectively a user defined conversion operation in C++.  So it reads as `RECT(from: rect)`.

Sources/UI/View.swift
farzonl: ah, I see b\c the spec says "If nWidth is set to CW_USEDEFAULT, the system ignores nHeight."  But I should probably also check for CW_USEDEFAULT on x as well.
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowa

for completeness though I can check all 4 if you like through a computed property on the Rect extension.

compnerd: Hmm, I think that a comment or just the equality check against `.default` would be fine.

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -29,22 +29,26 @@
 
 import WinSDK
 
-public extension Rect {
-  static let `default`: Rect =
+internal extension Rect {
+  public static let `default`: Rect =
```
compnerd: Is there something that is forcing this to be public?  Why not let this be internal (that is, users shouldn't use this magic value).

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -29,22 +29,26 @@
...
-  
-  var getRECT: RECT {
-      return RECT(left:   Int32(self.x), 
-                  top:    Int32(self.y),
-                  right:  Int32(self.width),
-                  bottom: Int32(self.height))
+}
+
+public extension Rect {
+  init(rect : RECT) {
```
compnerd: `init(from rect: RECT)` please.

Sources/UI/View.swift
```diff
@@ -45,10 +45,10 @@ public class View {
     self.style = style
 
     self.frame = frame
-    if (self.frame.width != Double(CW_USEDEFAULT)) {
-      var r = RECT(rect: self.frame);
+    if (!self.frame.isDefault) {
```
compnerd: The style for Swift is that there are no parenthesis with the the `if`.

```
  if self.frame != .default {
```

Sources/Support/Rect+UIExtensions.swift
farzonl: yes. ..\Sources\UI\Button.swift:71:39: error: static property 'default' is internal and cannot be referenced from a default argument value
  public override init(frame: Rect = .default, `class`: WindowClass = Button.class,

Sources/UI/View.swift
farzonl: I didn't want to do an equality check against .default incase someone were to make a Rect where just x or width is set to CW_USEDEFAULT

Sources/UI/View.swift
farzonl: ah ok i'll change this.

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -34,3 +34,28 @@ internal extension Rect {
...
+    return ( self.x == Double(CW_USEDEFAULT) || 
+             self.y == Double(CW_USEDEFAULT) ||
+             self.width == Double(CW_USEDEFAULT) ||
+             self.height == Double(CW_USEDEFAULT) )
+  }
+}
+
+public extension RECT {
+  init(from : Rect) {
+    self.init(left:   Int32(from.x),
```
compnerd: Could you please bin-pack the arguments?

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -34,3 +34,28 @@ internal extension Rect {
...
+
+public extension Rect {
+  init(from : RECT) {
+    self.x = Double(from.left)
+    self.y = Double(from.top)
+    self.width  = Double(from.right - from.left)
+    self.height = Double(from.bottom - from.top)
+  }
+
+  var isDefault : Bool {
```
compnerd: This doesn't really say if the value is entirely the default.  Perhaps a different name is more appropriate?  `isAnyPointDefault`?

Sources/Support/Rect+UIExtensions.swift
farzonl: is there a character limit I should abide by (80, 120, or other)?

Sources/Support/Rect+UIExtensions.swift
```diff
@@ -34,3 +34,25 @@ internal extension Rect {
...
+  }
+
+  var isAnyPointDefault: Bool {
+    return self.x == Double(CW_USEDEFAULT) || self.y == Double(CW_USEDEFAULT) ||
+           self.width == Double(CW_USEDEFAULT) || 
+           self.height == Double(CW_USEDEFAULT)
+  }
+}
+
+public extension RECT {
```
compnerd: Not sure if we really need this to be public.  The `Rect` type is a type we vend, `RECT` is the type that Microsoft vends.  I think that if we can avoid exposing `RECT` to users of Swift/Win32, it would be much nicer (avoids having to deal with type conversions upon porting).

compnerd:
@farzonl, would you mind squashing the commits into a single commit please?

PR #10, opened 2019-10-25
replace static class instances with factory methods

farzonl:
1. Idea here is for each view to have its own WindowClass instance. 
2. change the default behavior to redraw


farzonl:
This PR is about 4 commits back. There are build failures in later commits so I am intentionally not resolving the merge conflict until there are fixes available.

Issue #25, opened 2020-04-21
error: use of unresolved identifier 'DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2'


atirit:
I'm getting this error after running the following commands:

    cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Release . -D CMAKE_Swift_FLAGS="-sdk %SDK% -I %SDK%/usr/lib/swift -L %SDK%/usr/lib/swift/%OS%"
    ninja -C build SwiftWin32 HelloSwift

I've run `curl -sL "https://raw.githubusercontent.com/apple/swift/master/stdlib/public/Platform/winsdk.modulemap" -o "%UniversalCRTSdkDir%\Include\%UCRTVersion%\um\module.modulemap"` in an administrator Developer prompt and it made no difference.

Is there a different/newer/older modulemap I need to clone?

compnerd:
Bleh, this is a compatibility issue.  `DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2` doesn't get pulled in correctly from the clang importer.  We should provide a definition for it in `WinSDK+Extensions.swift` for compatibility.  Thanks for the report!

compnerd:
@atirit would you mind testing #26?  (CI currently is setup for 5.3)

atirit:
Sure! I'll do my best.

atirit:
I would get this by going through the Windows.md guide again, right?

compnerd:
@atirit no, that is far overkill.  I would say just apply the change (#26) to your repository and just build swift-win32.

atirit:
Gotcha.

atirit:
Adding the following to `Application.swift`:
```swift
#if swift(>=5.3)
#else
// windef.h
internal let DPI_AWARENESS_CONTEXT_UNAWARE: DPI_AWARENESS_CONTEXT =
    DPI_AWARENESS_CONTEXT(bitPattern: -1)!
internal let DPI_AWARENESS_CONTEXT_SYSTEM_AWARE: DPI_AWARENESS_CONTEXT =
    DPI_AWARENESS_CONTEXT(bitPattern: -2)!
internal let DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE: DPI_AWARENESS_CONTEXT =
    DPI_AWARENESS_CONTEXT(bitPattern: -3)!
[...]

PR #26, opened 2020-04-21
Support: add `DPI_AWARENESS_CONTEXT` constants

compnerd:
These are not imported properly by the clang importer.  WinSDK provides
the constants in 5.3+.  Provide definitions for compatibility with 5.2

Issue: #25


atirit:
The build failed; I don't think it likes `<=` as an operator for version numbers.

PR #40, opened 2020-06-10
Application: add hotkeys for rotation

compnerd:
This adds some hotkeys to emulate the ability to rotate applications on
platforms where rotation is not supported (e.g. desktops).

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-win32/40)


compnerd:
I dont think that I want to merge this right now.

PR #124, opened 2020-08-22
UI: extract the Windows specific bits of Color

compnerd:
Make Color more generic by externalising the Windows specific handling
of color where possible.


compnerd:
This is clearly not really ready yet, so closing the PR for now.

PR #158, opened 2020-09-27
UI: allow disabling minimize/maximize buttons

egorzhdan:
This PR adds `isMinimizable` & `isMaximizable` properties to the `Window` class.

### DIFF
--- a/Sources/UI/Window.swift
+++ b/Sources/UI/Window.swift
@@ -114,3 +114,15 @@ extension Window.Level {
   public static let statusBar: Window.Level = Window.Level(rawValue: 1000.0)
   public static let alert: Window.Level = Window.Level(rawValue: 2000.0)
 }
+
+extension Window {
+  public var isMaximizable: Bool {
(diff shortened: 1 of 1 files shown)


compnerd:
Would you mind splitting up the two commits into separate PRs?  I think that the style extraction is a great idea, but I would prefer that it be named `GWL_STYLE`.  I think for the minimize/maximize thing, at the very least, Id like to extract them out into an extension for the time being.

egorzhdan:
Extracted the properties into an extension. Had to add a little setup in the initializer to make it work, I don't know if that's okay considering this isn't a long-term solution.

compnerd (commented):
I think that adjusting the initializer is fine, but I just would like to understand it.  These were already set were they not?  Without the caption and system menu, the minimize and maximize buttons have no where to go, so that makes sense.  Perhaps a comment would be nice as well.

egorzhdan:
Wow, you're right, `WS_CAPTION` and `WS_SYSMENU` were already set in the `WindowStyle` initializer above, adding them again was meaningless. Removed it, now the change lies entirely in an extension.

PR #172, opened 2020-09-29
Extract `View.gwlStyle` property

egorzhdan:
Extracted from https://github.com/compnerd/swift-win32/pull/158.

I renamed the property into `gwlStyle` instead of `GWL_STYLE` to avoid clashing with the WinSDK constant, @compnerd please let me know if you still want it to be named in caps.

### DIFF
--- a/Sources/UI/Stepper.swift
+++ b/Sources/UI/Stepper.swift
@@ -77,12 +77,8 @@ public class Stepper: Control {
   public var isContinuous: Bool { fatalError("not yet implemented") }
   public var autorepeat: Bool { fatalError("not yet implemented") }
   public var wraps: Bool {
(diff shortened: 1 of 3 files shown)


compnerd:
I think that it is okay to clash with the Windows SDK constant as once we have the `GWL_STYLE` property, no one should be going directly for the style value, but rather using the property accessor.  Because this is also limited to internal scope, no one using the library would ever be impacted.

egorzhdan:
> no one should be going directly for the style value, but rather using the property accessor

Yeap, but it's kinda easy to accidentally skip `self.` and use the WinSDK constant instead of the property.

> Because this is also limited to internal scope, no one using the library would ever be impacted

Except if they use `@testable import` 🙂 but I guess there's no good reason to do that now.

Anyway, fixed the naming to the one you prefer.

compnerd:
This looks great!  Thank you!\

PR #173, opened 2020-09-30
Properly install/remove hooks

GunGraveKoga:
Calling SetWindowsHookExW with 0 as thread id will load SwiftWin32.dll in all GUI processes, but UnhookWindowsHookEx does not guarantie to unload that library at all.
In my case, actualy, I got high CPU usage on many processes like Explorer.exe, and 100% CPU usage on user session after application (Calculator.exe) termination.
Better to pass the id of the main thread to SetWindowsHookExW and set hooks for the application that calls ApplicationMain

### DIFF
--- a/Sources/Application/ApplicationMain.swift
+++ b/Sources/Application/ApplicationMain.swift
@@ -108,14 +108,16 @@ public func ApplicationMain(_ argc: Int32,
     log.error("GetModuleHandleExW: \(GetLastError())")
   }
 
+  let dwThreadId = GetCurrentThreadId()
+
   let hWindowProcedureHook: HHOOK? =
-      SetWindowsHookExW(WH_CALLWNDPROC, pApplicationWindowProc, hSwiftWin32, 0)
+      SetWindowsHookExW(WH_CALLWNDPROC, pApplicationWindowProc, hSwiftWin32, dwThreadId)
   if hWindowProcedureHook == nil {
     log.error("SetWindowsHookExW(WH_CALLWNDPROC): \(GetLastError())")
   }
 
   let hMessageProcedureHook: HHOOK? =
-      SetWindowsHookExW(WH_GETMESSAGE, pApplicationMessageProc, hSwiftWin32, 0)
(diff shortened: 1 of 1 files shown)


compnerd:
Hi @GunGraveKoga, thank you for tracking this down!

Sources/Application/ApplicationMain.swift
```diff
@@ -92,15 +92,26 @@ public func ApplicationMain(_ argc: Int32,
     log.error("GetModuleHandleExW: \(GetLastError())")
   }
 
+  let dwThreadId = GetCurrentThreadId()
+
   let hWindowProcedureHook: HHOOK? =
-      SetWindowsHookExW(WH_CALLWNDPROC, pApplicationWindowProc, hSwiftWin32, 0)
+      SetWindowsHookExW(WH_CALLWNDPROC, pApplicationWindowProc, hSwiftWin32, dwThreadId)
```
compnerd, line 98: Is there another need for the thread or is there any checking that needs to be done?  I think it might be nice to just inline the `GetCurrentThreadId` into the call to `SetWindowsHookExW`.  Would you mind adding a comment about this while you are at it?  I think that its very easy to overlook that, so a little comment to remind the reader is useful.

Sources/Application/ApplicationMain.swift
```diff
@@ -92,15 +92,26 @@ public func ApplicationMain(_ argc: Int32,
...
   if hWindowProcedureHook == nil {
     log.error("SetWindowsHookExW(WH_CALLWNDPROC): \(GetLastError())")
   }
 
+  func removeHooks() {
```
compnerd, line 103: Perhaps we can just use a `defer` block to perform the de-registration once we know that the hook has been installed successfully?

GunGraveKoga: Done

Sources/Application/ApplicationMain.swift
GunGraveKoga: I used the dwThreadId variable for multiple calls to SetWindowsHookEx. This is not needed now.
Done

compnerd:
Thanks!

compnerd (approved):

PR #179, opened 2020-10-03
Improve Application main loop

GunGraveKoga:
These changes allow us to use the RunLoop timers/sources and/or the Dispatch queue API on the main queue.
In future we should use CFRunLoop API to handle Win32 messages
_CFRunLoopSetWindowsMessageQueueMas and _CFRunLoopSetWindowsMessageQueueHandler wrapped in Swift style API

### DIFF
--- a/Sources/Application/ApplicationMain.swift
+++ b/Sources/Application/ApplicationMain.swift
@@ -151,24 +151,34 @@ public func ApplicationMain(_ argc: Int32,
   var nExitCode: Int32 = EXIT_SUCCESS
 
   mainLoop: while true {
+    // Process all messages in thread's message queue
+    // For GUI applications UI events must have high priority
     while PeekMessageW(&msg, nil, 0, 0, UINT(PM_REMOVE)) {
       guard msg.message != UINT(WM_QUIT) else {
+        // Handle WM_QUIT message, set application's exit code and terminate main loop
         nExitCode = Int32(msg.wParam)
         break mainLoop
       }
-
+      // Dispatch received message
       TranslateMessage(&msg)
       DispatchMessageW(&msg)
     }
 
     var limitDate: Date? = nil
     repeat {
+      // Execute Foundation.RunLoop once and determine the next time the timer fires
+      // At this point handles all Foundation.RunLoop timers, sources and Dispatch.DispatchQueue.main tasks
       limitDate = mainRunLoop.limitDate(forMode: .default)
+      // If Foundation.RunLoop doesn't contain any timer or timer should not be running right now
+      // we interrupt the current loop, otherwise go to the next iteration
       guard let limitDate = limitDate, limitDate.timeIntervalSinceNow <= 0 else {
         break
       }
     } while true
-
+    // Yields control to other threads
+    // If Foundation.RunLoop contain a timer to execute, we wait untill a new message is placed in thread's message queue
+    // or the timer must be fired, otherwise we proceed to the next iteration of mainLoop, using 0 as the wait timeout.
+    // WaitMessage(:) declared in WinSDK+Extensions.swift
     _ = WaitMessage(DWORD(limitDate?.timeIntervalSinceNow ?? 0 * 1000))
   }
 
--- a/Sources/Application/ApplicationMain.swift
+++ b/Sources/Application/ApplicationMain.swift
@@ -145,14 +145,36 @@ public func ApplicationMain(_ argc: Int32,
   }
 
   var msg: MSG = MSG()
-  while GetMessageW(&msg, nil, 0, 0) > 0 {
-    TranslateMessage(&msg)
-    DispatchMessageW(&msg)
+
+  let mainRunLoop = RunLoop.current
+
+  var exitCode: Int32 = EXIT_SUCCESS
+
+  mainLoop: while true {
+    while PeekMessageW(&msg, nil, 0, 0, UINT(PM_REMOVE)) {
+      guard msg.message != UINT(WM_QUIT) else {
+        exitCode = Int32(msg.wParam)
+        break mainLoop
+      }
+
+      TranslateMessage(&msg)
+      DispatchMessageW(&msg)
+    }
+
+    var limitDate: Date? = nil
+    repeat {
+      limitDate = mainRunLoop.limitDate(forMode: .default)
+      guard let limitDate = limitDate, limitDate.timeIntervalSinceNow <= 0 else {
+        break
+      }
+    } while true
+
+    _ = WaitMessage(DWORD(limitDate?.timeIntervalSinceNow ?? 0 * 1000))
   }
 
   Application.shared.delegate?.applicationWillTerminate(Application.shared)
 
-  return EXIT_SUCCESS
+  return exitCode
 }
 
 extension ApplicationDelegate {
--- a/Sources/Application/ApplicationMain.swift
+++ b/Sources/Application/ApplicationMain.swift
@@ -148,12 +148,12 @@ public func ApplicationMain(_ argc: Int32,
 
   let mainRunLoop = RunLoop.current
 
-  var exitCode: Int32 = EXIT_SUCCESS
+  var nExitCode: Int32 = EXIT_SUCCESS
 
   mainLoop: while true {
     while PeekMessageW(&msg, nil, 0, 0, UINT(PM_REMOVE)) {
       guard msg.message != UINT(WM_QUIT) else {
-        exitCode = Int32(msg.wParam)
+        nExitCode = Int32(msg.wParam)
         break mainLoop
       }
 
@@ -174,7 +174,7 @@ public func ApplicationMain(_ argc: Int32,
 
   Application.shared.delegate?.applicationWillTerminate(Application.shared)
 
-  return exitCode
+  return nExitCode
 }
 
 extension ApplicationDelegate {
--- a/Sources/Application/ApplicationMain.swift
+++ b/Sources/Application/ApplicationMain.swift
@@ -39,6 +39,16 @@ private let pApplicationWindowProc: HOOKPROC = { (nCode: Int32, wParam: WPARAM, 
   return CallNextHookEx(nil, nCode, wParam, lParam)
 }
 
+// Wait next message with timeout
+private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
+  let uIDEvent = WinSDK.SetTimer(nil, 0, dwMilliseconds, nil)
+  defer {
+    WinSDK.KillTimer(nil, uIDEvent)
+  }
+  // returned when a new message is placed in thread's message queue or timer expires
+  return WinSDK.WaitMessage()
+}
+
 @discardableResult
 public func ApplicationMain(_ argc: Int32,
                             _ argv: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>,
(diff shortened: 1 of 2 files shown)


compnerd:
@GunGraveKoga - thank you for exploring this, I've wanted the runloop integration for quite some time, but haven't been able to implement the dispatch runloop integration yet.  I'm going to have to ask for a bit more work on this and a lot more explanation, but I think that this is definitely the right direction and would love to get this integrated.  One thing though - I don't think that the `CFRunLoop` API is possible to use - it is unavailable from Foundation.

Sources/Application/ApplicationMain.swift
```diff
@@ -145,14 +145,36 @@ public func ApplicationMain(_ argc: Int32,
...
-  while GetMessageW(&msg, nil, 0, 0) > 0 {
-    TranslateMessage(&msg)
-    DispatchMessageW(&msg)
+
+  let mainRunLoop = RunLoop.current
+
+  var exitCode: Int32 = EXIT_SUCCESS
+
+  mainLoop: while true {
+    while PeekMessageW(&msg, nil, 0, 0, UINT(PM_REMOVE)) {
```
compnerd, line 154: Ah, I see, we need to use `PeekMessageW` so that we do not dequeue the message from the loop which would cause problems with `RunLoop`?  Why can we not just use `RunLoop` to run the message loop rather than having both the `RunLoop` from Foundation and a custom runloop here?

Sources/Application/ApplicationMain.swift
```diff
@@ -145,14 +145,36 @@ public func ApplicationMain(_ argc: Int32,
   }
 
   var msg: MSG = MSG()
-  while GetMessageW(&msg, nil, 0, 0) > 0 {
-    TranslateMessage(&msg)
-    DispatchMessageW(&msg)
+
+  let mainRunLoop = RunLoop.current
+
+  var exitCode: Int32 = EXIT_SUCCESS
```
compnerd, line 151: For better or worse, the codebase uses the Windows style and Hungarian notation :-(
```suggestion
  var dwExitCode: DWORD = EXIT_SUCCESS
```

Sources/Application/ApplicationMain.swift
```diff
@@ -145,14 +145,36 @@ public func ApplicationMain(_ argc: Int32,
...
+
+    var limitDate: Date? = nil
+    repeat {
+      limitDate = mainRunLoop.limitDate(forMode: .default)
+      guard let limitDate = limitDate, limitDate.timeIntervalSinceNow <= 0 else {
+        break
+      }
+    } while true
+
+    _ = WaitMessage(DWORD(limitDate?.timeIntervalSinceNow ?? 0 * 1000))
```
compnerd, line 172: Would just directly using `RunLoop` help avoid this complexity?  Having the entirety of the message loop being pumped by `RunLoop` seems so much better and would allow us to have the dispatch queues be backed by the same runloop.

adierking:
If it helps, here's some C code I wrote a while ago to drive a Windows message loop with Dispatch without CF - https://gist.github.com/adierking/4a12504ddcb5348ce1ab30d5e5897634

This makes use of the internal `4CF` APIs though :(

GunGraveKoga:
On Windows and Linux `RunLoop` uses `CFRunLoop` under the hood, so it's just `CFRunLoop` wrapped in `RunLoop` API. 
I think on Windows we should have a `RunLoop` method to handle Windows messages, like this:

```swift
#if os(Windows)
open func acceptInput(forMode mode: RunLoop.Mode, before limitDate: Date, withMask mask: UInt32, handler:  @convention(c) () -> Void) {
  if _cfRunLoop !== CFRunLoopGetCurrent() {
    return
  }
  let oldMask = _CFRunLoopGetWindowsMessageQueueMask(_cfRunLoop, mode._cfObject)
  defer {
    _CFRunLoopSetWindowsMessageQueueMask(_cfRunLoop, oldMask, mode._cfObject)
  }

  let oldHandler = _CFRunLoopGetWindowsMessageQueueHandler(_cfRunLoop, mode._cfObject)
  defer {
    _CFRunLoopSetWindowsMessageQueueHandler(_cfRunLoop, mode._cfObject, oldHandler)
  }

  _CFRunLoopSetWindowsMessageQueueMask(_cfRunLoop, mask, mode._cfObject)
 _CFRunLoopSetWindowsMessageQueueHandler(_cfRunLoop, mode._cfObject, handler)

  CFRunLoopRunInMode(mode._cfObject, limitDate.timeIntervalSinceReferenceDate - CFAbsoluteTimeGetCurrent(), true)
}
#endif
```

Sources/Application/ApplicationMain.swift
GunGraveKoga: Cos `RunLoop` cannot process Windows messages at the moment.
We can use a `Timer` to process Windows messages, but this method causes high CPU usage.
The main idea behind my implementation:
- peek, dequeue and dispatch all available messages
- process all `RunLoop` timers and sources that should be executed at this point of time
- wait until a new message is queued or until the next `RunLoop` timer should starts (yield calling thread)

compnerd:
Requiring changes to foundation makes this much less preferable as that would need to be something that upstream accepts.  I think that the way to handle this is to use a custom CFRunLoopSource and attach that to the main RunLoop.

Sources/Application/ApplicationMain.swift
GunGraveKoga: ApplicationMain returns Int32, and I dont think we should convert msg.wParam (UINT) ot DWORD  and than convert exit code (DWORD aka UInt32) to Int32.
I renamed the variable to nExitCode

GunGraveKoga:
I don't see how CFRunLoopSource can solve problem of Windows messages handling

Sources/Application/ApplicationMain.swift
GunGraveKoga: I use `WaitMessage` with timeout to yield main thread and prevent high CPU usage.
If `RunLoop` contains timer which need to execute, `WaitMessage` will wait timer fire moment or new Windows message was queued, in other case, `WaitMessage` returns immediately.
In any case, we should peek all messages, befor `RunLoop` timers/sources processing

Sources/Application/ApplicationMain.swift
compnerd: Ah, agreed.  Thanks.

compnerd:
The idea is that you would create a CFRunLoopSource and add that source to the main runloop.  Then you just pump that main runloop, which will process both the windows message loop from the source you added and the dispatch runloops.

Sources/Application/ApplicationMain.swift
GunGraveKoga: Should I add more comments to explain my main loop implementation?

compnerd: Yes please.  Im not sure I understand why we need to peek and dequeue the operation rather than simply dequeue and dispatch.  The driving the dispatch main loop intertwined with the windows main loop makes sense.

compnerd: I think that the `CFRunLoopSource` solution is still better, but, a slightly kludgy solution to get dispatch pumped is more valuable than maintaining the cleanliness right now.

GunGraveKoga:
If pass QS_ALLEVENTS or QS_ALLINPUT as mask to _CFRunLoopSetWindowsMessageQueueMask for RunLoop, RunLoop will process Windows messages.
By default CFRunLoop use PeekMessage as default Windows message queue handler.
In any case some thing should call CFRunLoopSourceSignal to fire it.

compnerd:
> In any case some thing should call CFRunLoopSourceSignal to fire source.

Yes, this is exactly the point that I was making earlier about having a custom source :). I'll upload an experimental change which doesn't work yet due to not being able to get an observer to signal the source working.  It might help better explain what I was suggesting.

compnerd:
@GunGraveKoga what I was suggesting with the RunLoopSource was something along the lines of https://github.com/compnerd/swift-win32/pull/185 which still needs a bit more work, but should allow pumping all the loops at once.

Sources/Application/ApplicationMain.swift
GunGraveKoga: Added additional comments explaining the main loop implementation

Sources/Application/ApplicationMain.swift
```diff
@@ -145,14 +145,46 @@ public func ApplicationMain(_ argc: Int32,
...
-    TranslateMessage(&msg)
-    DispatchMessageW(&msg)
+
+  let mainRunLoop = RunLoop.current
+
+  var nExitCode: Int32 = EXIT_SUCCESS
+
+  mainLoop: while true {
+    // Process all messages in thread's message queue
+    // For GUI applications UI events must have high priority
```
compnerd, line 155: Comments in general should be complete sentences.

Sources/Support/WinSDK+Extensions.swift
```diff
@@ -75,3 +75,13 @@ func EnableMenuItem(_ hMenu: HMENU, _ uIDEnableItem: UINT, _ uEnable: UINT)
                     to: ((HMENU?, UINT, UINT) -> CInt).self)
   return pfnEnableMenuItem(hMenu, uIDEnableItem, uEnable)
 }
+
+// Wait next message with timeout
+func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
```
compnerd, line 80: This is probably better to be a `private` function in `ApplicationMain.swift` I think.  This is not something that I would think would be likely to go into the WinSDK module in Swift.

Sources/Support/WinSDK+Extensions.swift
```diff
@@ -75,3 +75,13 @@ func EnableMenuItem(_ hMenu: HMENU, _ uIDEnableItem: UINT, _ uEnable: UINT)
...
+  let timerId = WinSDK.SetTimer(nil, 0, dwMilliseconds, nil)
```
compnerd, line 81: I think that `uIDEvent` is a better name as that follows the Win32 API naming scheme.

compnerd:
@GunGraveKoga - I think you've convinced me about the patch (modulo some minor tweaks that I think would be nice).  I still think that using a custom source will be way nicer to pump the loop, but we can do that later.

GunGraveKoga:
> @GunGraveKoga what I was suggesting with the RunLoopSource was something along the lines of #185 which still needs a bit more work, but should allow pumping all the loops at once.

Yes, it's better solution, but `Foundation` API for Swift 5.3 doesn't contains `_add` and `_observe` methods for `RunLoop`

Sources/Support/WinSDK+Extensions.swift
GunGraveKoga: Done

Sources/Support/WinSDK+Extensions.swift
GunGraveKoga: Done

Sources/Application/ApplicationMain.swift
GunGraveKoga: Done

compnerd:
True, though, I think that keeping to 5.3 might be difficult as more functionality is added.  The snapshots are really preferable to use I think.  Either way, getting dispatch integrated is valuable enough to actually go with this route until we can fix that.

I think for this particular patch, its been long enough that I can just fix it in a follow up, but, the code style for this project tries to stick to 80-columns.

GunGraveKoga:
An one more thing, for some reason, when I received the WM_TIMER message,
```swift
func GetMessageW(LPMSG?, HWND?, UINT, UINT) -> CInt
``` 
started returning negative value for any message and
```swift
while GetMessageW(&msg, nil, 0, 0) > 0
```
dose not work properly

compnerd:
Hmm, that is somewhat worrisome.  I think that we should understand why that would fail before we merge this :-(

GunGraveKoga:
`GetLastError()` is always 0
But if hooking code (SetWindowsHookExW) is commented, all works fine.
Again, if use `GetMessageW(LPMSG?, HWND?, UINT, UINT) -> Bool` all works fine too.

compnerd:
Interesting, I wonder if something is going wrong with the cast then.  But, since the `Bool` version, which is the native import works, I think that we can deal with that later.  Hopefully this wont actually break the example applications.

compnerd:
Ugh, this is causing the demo application to immediately exit for me :/  Were you able to run UICatalog @GunGraveKoga?

PR #189, opened 2020-10-06
Revert "Improve Application main loop"

compnerd:
This reverts commit 8706996e99610a0cf462501756a666986daca3ac.

With this change UICatalog immediately exits upon launch rather than
continuing to pump the main loop.  The application launch completes
successfully as the key window is rendered, but then terminates shortly
thereafter.  Revert the change until the mainloop can be pumped.


compnerd:
CC: @GunGraveKoga

PR #191, opened 2020-10-07
Improve Application main loop (fixed)

GunGraveKoga:
Same as https://github.com/compnerd/swift-win32/pull/179 with fixed Double to DWORD conversion


compnerd:
Thanks @GunGraveKoga!

PR #224, opened 2020-10-24
UI: add `SUBCLASSPROC` for `View`

egorzhdan:
Invoke `SetWindowSubclass` for all kinds of views in `View.init`, set the default `SUBCLASSPROC` for views which do not provide a custom implementation. This will allow `View` to handle popup menu events like `WM_CONTEXTMENU`.

### DIFF
--- a/Sources/UI/View.swift
+++ b/Sources/UI/View.swift
@@ -7,6 +7,18 @@
 
 import WinSDK
 
+private let SwiftViewProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
+  let view: View? = unsafeBitCast(dwRefData, to: AnyObject.self) as? View
+  switch uMsg {
+  case UINT(WM_CONTEXTMENU):
+    // TODO handle popup menu events
+    return 0
+  default:
+    break
+  }
+  return DefSubclassProc(hWnd, uMsg, wParam, lParam)
+}
+
 internal typealias WindowStyle = (base: DWORD, extended: DWORD)
 
 private func ClientToWindow(size: inout Size, for style: WindowStyle) {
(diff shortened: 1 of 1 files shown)


compnerd:
Hi @egorzhdan,

Thanks for the PR.  I think that having a single point where `WM_CONTEXTMENU` is handled is a good idea.  However, I don't know if we really need to subclass the window behind `View` to do that.  We could always dispatch to share the code.  If we have a conditional replacement of the window proc to avoid the upcast, we will end up with the dispatch in any case.  Is there something that I am overlooking?

Alternatively, we could have multiple subclasses of the window itself.

https://docs.microsoft.com/en-us/windows/win32/controls/subclassing-overview explains that (the version 6 approach is in use here).

egorzhdan:
@compnerd thanks for the hint!
The idea behind moving `SetWindowSubclass` invocation to `View` was to make sure we only register one subclass for each control, taking into account that some controls declare a custom `SUBCLASSPROC ` and others don't. But it probably doesn't make much sense, since having multiple subclasses is allowed and works fine.
I think the easiest solution is probably be to register one more subclass in `View` for handling popup menu events, and call `DefSubclassProc` in all the subclasses without delegation to `SwiftViewProc` (is it what you meant by using dispatch?)
What do you think?

Sources/UI/View.swift
```diff
@@ -150,6 +162,9 @@ public class View: Responder {
 
     super.init()
 
+    SetWindowSubclass(self.hWnd, SwiftViewProc, UINT_PTR.max,
```
compnerd, line 165: I've been using an explicit `_ =` assignment wherever there is a return value that I am not checking.  It just makes it much more clear that the error conditions aren't being checked - as they really should be.  Would you mind adding that in?

compnerd (commented):
Yeah, this is much more inline with what came to mind when I saw the change.  The idea is that in the case that we need to invoke the handler in the view itself, we can always expose that *internally* and have the derived type call the base type's handler.

Minor quibble on the error handling, but looking forward to seeing this come together.

Sources/UI/View.swift
egorzhdan: Ah, sure, fixed that.

PR #246, opened 2020-10-30
UI: support showing basic context menus

egorzhdan:
This change adds support for basic context menus (attached to a `View` instance) and window title menus. For both types of menus submenus are supported.

### DIFF
--- a/Sources/SwiftWin32/UI/Menu.swift
+++ b/Sources/SwiftWin32/UI/Menu.swift
@@ -5,6 +5,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  **/
 
+import WinSDK
+
 extension Menu {
   /// Constants for identifying an application's standard menus.
   public struct Identifier: Equatable, Hashable, RawRepresentable {
@@ -275,12 +277,30 @@ extension Menu.Options {
 /// A container for grouping related menu elements in an application menu or
 /// contextual menu.
 public class Menu: MenuElement {
+  internal let children: [MenuElement]
+
   /// Creating a Menu Object
 
   /// Creates a new menu with the specified values.
   public init(title: String = "", image: Image? = nil,
               identifier: Menu.Identifier? = nil, options: Menu.Options = [],
               children: [MenuElement] = []) {
+    self.children = children
     super.init(title: title, image: image)
+  }
+}
+
+internal struct Win32Menu {
+  internal let hMenu: MenuHandle
+  private let children: [Win32MenuElement]
+
+  internal init(_ hMenu: MenuHandle, children: [MenuElement]) {
+    self.hMenu = hMenu
+    self.children = children.map { Win32MenuElement.of($0) }
+    for (index, child) in self.children.enumerated() {
+      InsertMenuItemW(hMenu.value,
+                      UINT(index), true,
+                      &(child.info))
+    }
   }
 }
--- a/Sources/SwiftWin32/UI/MenuElement.swift
+++ b/Sources/SwiftWin32/UI/MenuElement.swift
@@ -5,6 +5,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  **/
 
+import WinSDK
+
 extension MenuElement {
   /// Attributes that determine the style of the menu element.
   public struct Attributes: OptionSet {
@@ -65,5 +67,51 @@ public class MenuElement {
   public init(title: String, image: Image? = nil) {
     self.title = title
     self.image = image
+  }
+}
+
+internal class Win32MenuElement {
+  internal var info: MENUITEMINFOW
+  private var titleW: [WCHAR]
+  private let subMenu: Win32Menu?
+
+  private init(title: String, subMenu: Win32Menu?, fType: Int32) {
+    self.titleW = title.LPCWSTR
+    let titleSize = titleW.count
+
+    self.subMenu = subMenu
+    self.info = titleW.withUnsafeMutableBufferPointer {
+      MENUITEMINFOW(cbSize: UINT(MemoryLayout<MENUITEMINFOW>.size),
+                    fMask: UINT(MIIM_FTYPE | MIIM_STATE | MIIM_ID | MIIM_STRING | MIIM_SUBMENU | MIIM_DATA),
+                    fType: UINT(fType),
+                    fState: UINT(MFS_ENABLED),
+                    wID: UInt32.random(in: .min ... .max),
+                    hSubMenu: subMenu?.hMenu.value,
+                    hbmpChecked: nil,
+                    hbmpUnchecked: nil,
+                    dwItemData: 0,
+                    dwTypeData: $0.baseAddress,
+                    cch: UINT(titleSize),
+                    hbmpItem: nil)
+    }
+  }
+
+  internal static func of(_ element: MenuElement) -> Win32MenuElement {
+    let subMenu: Win32Menu?
+    if let menu = element as? Menu {
+      subMenu = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
+                          children: menu.children)
+    } else {
+      subMenu = nil
+    }
+    return Win32MenuElement(title: element.title, subMenu: subMenu, fType: MFT_STRING)
+  }
+
+  internal static var separator: Win32MenuElement {
+    Win32MenuElement(title: "", subMenu: nil, fType: MFT_SEPARATOR)
+  }
+
+  internal var isSeparator: Bool {
+    info.fType == MFT_SEPARATOR
   }
 }
--- a/Sources/SwiftWin32/UI/View.swift
+++ b/Sources/SwiftWin32/UI/View.swift
@@ -11,8 +11,27 @@ private let SwiftViewProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubc
   let view: View? = unsafeBitCast(dwRefData, to: AnyObject.self) as? View
   switch uMsg {
   case UINT(WM_CONTEXTMENU):
-    // TODO handle popup menu events
+    guard let view = view,
+          let menuInteraction = view.interactions.first(where: { $0 is ContextMenuInteraction })
+              as? ContextMenuInteraction else { break }
+    let x = Int16(truncatingIfNeeded: lParam)
+    let y = Int16(truncatingIfNeeded: lParam >> 16)
+    let point = Point(x: Int(x), y: Int(y))
+    let menuConfiguration = menuInteraction.delegate?.contextMenuInteraction(menuInteraction,
+                                                                             configurationForMenuAtLocation: point)
+    if let menu = menuConfiguration?.provideActions() {
+      view.win32ContextMenu = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
+                                        children: menu.children)
+    } else {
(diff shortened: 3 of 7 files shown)


compnerd:
discussed offline - the API surface is likely going to revert to something that was there previously

egorzhdan:
(only rebased for now, will address the feedback a bit later)

Sources/SwiftWin32/UI/MenuElement.swift
```diff
@@ -55,15 +57,60 @@ public class MenuElement {
...
+  public let image: Image?
 
   /// Creating a Menu Element
 
-  internal init(title: String, image: Image?) {
+  public init(title: String, image: Image? = nil) {
     self.title = title
     self.image = image
   }
 }
```
compnerd, line 71: Sorry about that.  Would you mind splitting out the interface changes to `MenuElement` into a separate change that precedes the implementation?  This is fixing the interface definition that I had gotten incorrect.

egorzhdan: Okay, extracted these changes into https://github.com/compnerd/swift-win32/pull/294

Sources/SwiftWin32/UI/Menu.swift
```diff
@@ -275,12 +277,42 @@ extension Menu.Options {
...
   public init(title: String = "", image: Image? = nil,
               identifier: Menu.Identifier? = nil, options: Menu.Options = [],
               children: [MenuElement] = []) {
+    self.children = children
     super.init(title: title, image: image)
   }
 }
+
+internal class WrappedMenu {
+  internal let hMenu: HMENU
```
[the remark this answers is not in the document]
egorzhdan: Wow, I haven't noticed there's an existing mechanism for this kind of things. Fixed.

Examples/UICatalog/UICatalog.swift
```diff
@@ -110,6 +110,19 @@ final class UICatalog: ApplicationDelegate, SceneDelegate {
     window.addSubview(self.tableview)
     window.addSubview(self.imageview)
 
+    window.rootViewController?.titleMenu = Menu(children: [
+      Menu(title: "Foo", children: [
```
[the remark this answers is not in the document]
egorzhdan: Yeap, sounds good, replaced the menu with "Application", since it's right under the window title which says "UICatalog".

If I understand the API design correctly, the logic for suggested menu items would belong somewhere in `ContextMenuConfiguration`, but I haven't really touched it yet.

Examples/UICatalog/UICatalog.swift
```diff
@@ -110,6 +110,19 @@ final class UICatalog: ApplicationDelegate, SceneDelegate {
...
+        Menu(title: "Another Submenu", children: [
+          MenuElement(title: "Deep Item")
+        ])
+      ]),
+      Menu(title: "Bar", children: [
+        MenuElement(title: "Item Without a Submenu")
+      ]),
+    ])
+
+    self.progress.addInteraction(ContextMenuInteraction(delegate: self))
```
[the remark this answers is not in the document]
egorzhdan: Turns out the menu logic doesn't actually work for `Label`s – the `WM_CONTEXTMENU` events are dispatched to a `Window` containing the label instead of the label itself.
We'll probably need a `LabelProxy` to be able to handle the events properly, I'll add it in a separate PR.

Examples/UICatalog/UICatalog.swift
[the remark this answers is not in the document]
compnerd: Actually, I realized that you are looking at context menus, not application menus.  In such a case, perhaps the application doesn't make sense :-(.  But creating a new UI element to attach the context menu makes sense (perhaps a TextField?)

Examples/UICatalog/UICatalog.swift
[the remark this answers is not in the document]
compnerd: Adding the proxy sounds good to me, and as a separate PR sounds perfect!

Sources/SwiftWin32/UI/Menu.swift
```diff
@@ -275,12 +277,38 @@ extension Menu.Options {
...
+    self.children = children
     super.init(title: title, image: image)
   }
 }
+
+internal class Win32Menu {
+  internal let hMenu: MenuHandle
+  private let children: [Win32MenuElement]
+
+  private init(hMenu: HMENU, children: [MenuElement]) {
```
compnerd, line 297: Why not just take a `MenuHandle` instead?  Then, rather than the constructors below, we could just pass in the `MenuHandle` constructed with the correct Win32 call.

```swift
_ = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
              children: children)
_ = Win32Menu(MenuHandle(owning: CreateMenu()),
              children: children)
```

Sources/SwiftWin32/UI/Menu.swift
```diff
@@ -275,12 +277,38 @@ extension Menu.Options {
...
   /// Creates a new menu with the specified values.
   public init(title: String = "", image: Image? = nil,
               identifier: Menu.Identifier? = nil, options: Menu.Options = [],
               children: [MenuElement] = []) {
+    self.children = children
     super.init(title: title, image: image)
   }
 }
+
+internal class Win32Menu {
```
compnerd, line 293: Are there reference semantics which are needed or can we get away with value semantics of `struct`?

Sources/SwiftWin32/UI/MenuElement.swift
```diff
@@ -67,3 +69,48 @@ public class MenuElement {
...
+    let subMenu: Win32Menu?
+    if let menu = element as? Menu {
+      subMenu = Win32Menu.ofContextMenu(menu)
+    } else {
+      subMenu = nil
+    }
+    return Win32MenuElement(title: element.title, subMenu: subMenu, fType: MFT_STRING)
+  }
+
+  internal static func ofSeparator() -> Win32MenuElement {
```
compnerd, line 109: Why not make this a computed property:

```swift
  internal static var separator: Win32MenuElement {
    Win32MenuElement(title: "", submenu: nil, fType: MFT_SEPARATOR)
  }
```

Sources/SwiftWin32/UI/MenuElement.swift
```diff
@@ -67,3 +69,48 @@ public class MenuElement {
     self.image = image
   }
 }
+
+internal class Win32MenuElement {
```
compnerd, line 73: Are there reference semantics that are needed, or can we use value semantics here?

Sources/SwiftWin32/UI/Menu.swift
egorzhdan: Okay, fixed

Sources/SwiftWin32/UI/Menu.swift
egorzhdan: Nope, fixed.

Sources/SwiftWin32/UI/MenuElement.swift
egorzhdan: Fixed

Sources/SwiftWin32/UI/MenuElement.swift
egorzhdan: We're passing a reference to `Win32MenuElement.info` as a parameter to `InsertMenuItemW` on `Menu.swift:303`, that won't compile if `Win32MenuElement` is a struct.

```
Menu.swift:303:23: error: cannot pass immutable value as inout argument: 'child' is a 'let' constant
                      &(child.info))
                      ^ ~~~~~
```

compnerd: Hmm, what if you make `children` a `var` instead?  I think that it might be possible to get away with that.

egorzhdan: Nope, that doesn't help, the same error occurs. That's probably because even though `children` is mutable, its elements still need `mutating` keyword for modifying operations.

Sources/SwiftWin32/UI/MenuElement.swift
```diff
@@ -67,3 +69,49 @@ public class MenuElement {
...
+    }
+  }
+
+  internal static func of(_ element: MenuElement) -> Win32MenuElement {
+    let subMenu: Win32Menu?
+    if let menu = element as? Menu {
+      subMenu = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
+                          children: menu.children)
+    } else {
+      subMenu = nil
```
compnerd, line 105: This is unnecessary, the declaration should be initialized to `nil` anyways.

Sources/SwiftWin32/UI/View.swift
```diff
@@ -11,8 +11,27 @@ private let SwiftViewProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubc
...
+    let x = Int16(truncatingIfNeeded: lParam)
+    let y = Int16(truncatingIfNeeded: lParam >> 16)
+    let point = Point(x: Int(x), y: Int(y))
+    let menuConfiguration = menuInteraction.delegate?.contextMenuInteraction(menuInteraction,
+                                                                             configurationForMenuAtLocation: point)
+    if let menu = menuConfiguration?.provideActions() {
+      view.win32ContextMenu = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
+                                        children: menu.children)
+    } else {
+      view.win32ContextMenu = nil
```
compnerd, line 26: This should not be need right?

Sources/SwiftWin32/UI/MenuElement.swift
egorzhdan: If I remove the `else` block, it stops compiling:
```
swift-win32\Sources\SwiftWin32\UI\MenuElement.swift:105:60: error: constant 'subMenu' used before being initialized
    return Win32MenuElement(title: element.title, subMenu: subMenu, fType: MFT_STRING)
                                                           ^
```

Sources/SwiftWin32/UI/View.swift
egorzhdan: This is executed if the menu was non-`nil` during previous `contextMenuInteraction` invocation, but is `nil` now and shouldn't be shown. The assignment also removes the reference to the outdated menu, allowing it to deinit if not referenced anywhere else.

Examples/UICatalog/UICatalog.swift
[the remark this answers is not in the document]
egorzhdan: Unfortunately adding the proxy didn't help, but wrapping the label in a view solves the issue. Sent https://github.com/compnerd/swift-win32/pull/306.

Examples/UICatalog/UICatalog.swift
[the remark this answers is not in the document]
egorzhdan: Oh, I've mixed up the menu types in my comment above, sorry. The `titleMenu` field is about window title menus, not context menus, so `ContextMenuConfiguration` doesn't apply here.

egorzhdan:
I think I've addressed all the problematic parts: removed window title menus (will add as a separate PR), and fixed the code issues.

compnerd (approved):
Thanks!

egorzhdan:
Just added the missing file to CMakeLists, the build should pass now.

PR #282, opened 2020-11-21
UICatalog: use the correct bundle when building with SwiftPM

egorzhdan:
SwiftPM generates a `Bundle.module` field, which should be used for accessing resources associated with the current target.
Using `Bundle.main` causes a runtime error in this case, since the requested image URL is `nil`.


compnerd (approved):
This is rather unfortunate :-(  Thanks for fixing this!

PR #291, opened 2020-11-26
Platform: sketch out an orientation monitor

compnerd:
Use the Windows Sensors API to monitor the system orientation to allow
for notification on orientation changes.

### DIFF
--- a/Sources/SwiftWin32/Platform/OrientationSensor.swift
+++ b/Sources/SwiftWin32/Platform/OrientationSensor.swift
@@ -1,0 +1,82 @@
+/**
+ * Copyright © 2020 Saleem Abdulrasool <compnerd@compnerd.org>
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ **/
+
+import WinSDK
+import SwiftCOM
+
+internal class OrientationSensorEventListener: SwiftCOM.ISensorEvents {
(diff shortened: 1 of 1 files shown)


Sources/SwiftWin32/Platform/OrientationSensor.swift
```diff
@@ -0,0 +1,82 @@
...
+ **/
+
+import WinSDK
+import SwiftCOM
+
+internal class OrientationSensorEventListener: SwiftCOM.ISensorEvents {
+  public func OnEvent(_ pSensor: SwiftCOM.ISensor, _ eventID: REFGUID,
+                      _ pEventData: SwiftCOM.IPortableDeviceValues) -> HRESULT {
+    switch eventID {
+    case SENSOR_EVENT_PROPERTY_CHANGED:
```
egorzhdan, line 15: This doesn't compile for me when building with SwiftPM:
```
error: expression pattern of type 'GUID' (aka '_GUID') cannot match values of type 'REFGUID' (aka 'UnsafePointer<_GUID>')
    case SENSOR_EVENT_PROPERTY_CHANGED:
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

@compnerd I think you might've forgotten to add this file to the CMakeLists, and because of that it isn't compiled on the CI 😉

compnerd: Ah, local changes - yes, this isn't integrated into the build (something I do really miss with SPM - explicit source lists)

PR #296, opened 2020-11-28
Support: hoist GUID equality into the Support layer

compnerd:
The ability to check for `GUID` equality is rather common on Windows.
Move the internal implementation from the `BatteryMonitor` into the
WinSDK extensions and make it available internally to the module.


compnerd:
CC: @egorzhdan

PR #305, opened 2020-12-04
Update swift snapshot

damuellen:
The build 2020-12-01 does not work for me, let's see if it runs here.


compnerd:
This was broken as a result of a change that went into Swift Package Manager.  See https://github.com/compnerd/swift-build/commit/70033af8184f836e4e145288f54e93e0420c115a.

PR #306, opened 2020-12-05
UI: add containing `View` for each `Label`

egorzhdan:
`WC_STATIC` control doesn't receive events like `WM_CONTEXTMENU`. To be able to handle those events, let's create containing views for `WC_STATIC` instances.

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/Label.swift
+++ b/Sources/SwiftWin32/Views and Controls/Label.swift
@@ -21,21 +21,61 @@ private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSub
 }
 
 public class Label: Control {
-  private static let `class`: WindowClass = WindowClass(named: WC_STATIC)
-  private static let style: WindowStyle = (base: WS_TABSTOP | DWORD(SS_NOTIFY), extended: 0)
+  private static let `class`: WindowClass =
+      WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Label")
+  private static let style: WindowStyle = (base: WS_TABSTOP, extended: 0)
 
-  public override var font: Font! {
-    get { return super.font }
-    set(value) { super.font = value }
+  private var staticHWnd: HWND!
+
+  public var text: String? {
+    get {
+      let szLength: Int32 = GetWindowTextLengthW(self.staticHWnd)
+      let buffer: [WCHAR] = Array<WCHAR>(unsafeUninitializedCapacity: Int(szLength) + 1) {
+        $1 = Int(GetWindowTextW(self.staticHWnd, $0.baseAddress!, CInt($0.count)))
+      }
+      return String(decodingCString: buffer, as: UTF16.self)
+    }
+    set(value) { _ = SetWindowTextW(self.staticHWnd, value?.wide) }
   }
 
-  @_Win32WindowText
-  public var text: String?
+  public override var font: Font! {
+    didSet {
+      SendMessageW(self.staticHWnd, UINT(WM_SETFONT),
+                   unsafeBitCast(self.font?.hFont.value, to: WPARAM.self),
+                   LPARAM(1))
+    }
+  }
+
+  public override var frame: Rect {
+    didSet {
+      let rect = GetRect(hWnd: self.hWnd)
+      _ = SetWindowPos(self.staticHWnd, nil,
+                       CInt(rect.origin.x), CInt(rect.origin.y),
+                       CInt(rect.size.width), CInt(rect.size.height),
+                       UINT(SWP_NOZORDER | SWP_FRAMECHANGED))
+    }
+  }
 
   public init(frame: Rect) {
     super.init(frame: frame, class: Label.class, style: Label.style)
     _ = SetWindowSubclass(hWnd, SwiftLabelProc, UINT_PTR(1),
                          unsafeBitCast(self as AnyObject, to: DWORD_PTR.self))
+
+    let rect = GetRect(hWnd: self.hWnd)
+    self.staticHWnd = CreateWindowExW(0, WC_STATIC.wide, nil, 0,
+                                      0, 0,
+                                      Int32(rect.size.width),
+                                      Int32(rect.size.height),
+                                      nil, nil, GetModuleHandleW(nil), nil)!
+
+    _ = SetWindowLongW(self.staticHWnd, WinSDK.GWL_STYLE, WS_CHILD)
+    _ = SetParent(self.staticHWnd, self.hWnd)
+
+    self.font = Font.systemFont(ofSize: Font.systemFontSize)
+  }
+
+  deinit {
+    DestroyWindow(self.staticHWnd)
   }
 
   // ContentSizeCategoryAdjusting
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -76,6 +76,19 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle) 
   rect = Rect(from: r)
 }
 
+internal func GetRect(hWnd: HWND) -> Rect {
+  var r: RECT = RECT()
+  if !GetClientRect(hWnd, &r) {
+    log.warning("GetClientRect: \(Error(win32: GetLastError()))")
+  }
+  _ = withUnsafeMutablePointer(to: &r) { [hWnd] in
+    $0.withMemoryRebound(to: POINT.self, capacity: 2) {
+      MapWindowPoints(hWnd, nil, $0, 2)
+    }
+  }
+  return Rect(from: r)
+}
+
 extension View {
   /// Options to specify how a view adjusts its content when its size changes.
   public enum ContentMode: Int {
@@ -241,16 +254,7 @@ public class View: Responder {
     // If `CW_USEDEFAULT` was used, query the actual allocated rect
     if frame.origin.x == Double(CW_USEDEFAULT) ||
        frame.size.width == Double(CW_USEDEFAULT) {
-      var r: RECT = RECT()
-      if !GetClientRect(self.hWnd, &r) {
-        log.warning("GetClientRect: \(Error(win32: GetLastError()))")
-      }
-      _ = withUnsafeMutablePointer(to: &r) { [hWnd = self.hWnd] in
-        $0.withMemoryRebound(to: POINT.self, capacity: 2) {
-          MapWindowPoints(hWnd, nil, $0, 2)
-        }
-      }
-      client = Rect(from: r)
+      client = GetRect(hWnd: self.hWnd)
     }
 
     // Scale window for DPI


compnerd:
Hmm, can we not just adjust the owner window for label and handle the event?

egorzhdan:
I think it would significantly complicate the `Window` event handling logic: we'd have to go through all the labels in all the components inside the window, find the labels that correspond to the coordinates of the mouse click, filter out the hidden ones, and find the topmost enabled label.
Similar logic would be required for other interactions to work, like drag & drop.

I believe it would be better to delegate this work to the OS, just like for all the other components.

compnerd:
Thinking more about this, is this really a special case?  Will this be needed for all the controls?  Can we avoid this by doing a custom draw?

egorzhdan:
> is this really a special case

I think so, it seems `WC_STATIC` is the only built-in control that doesn't receive all the click events. At least I couldn't find another such control.

> Will this be needed for all the controls?

Nope, this is only needed for `WC_STATIC`.
Actually I've just noticed that `ImageView` also relies on `WC_STATIC`, so it might need a similar workaround as `Label`.

> Can we avoid this by doing a custom draw?

Not sure... I'll try.

> The hit testing can be delegated to the OS

This would be great, this way there won't be the need to reimplement the system event dispatch logic. I don't really know how to do it though...


A similar issue will probably arise once drag-and-drop or other interactions are implemented – these static controls won't be receiving the events properly.

compnerd:
This would be good to get rebased and prepared to be merged, at least for the Label case.  We should figure out how to do this more generically at some point.

egorzhdan:
Just rebased this, and it still works for me – the context menus show up on a right-click on labels as expected.
I haven't tried to abstract this logic away from `Label` in this PR, but I think it would eventually be needed to support context menus for `Image`.

compnerd:
Hey @egorzhdan, sorry to do this to you, but could you rebase this on top of #434 as I think that is a pretty simple change and I was thinking about a bit more tweaking to this change?

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
 
 import WinSDK
 
-public class Label: Control {
+fileprivate class Win32Label: View {
   private static let `class`: WindowClass = WindowClass(named: WC_STATIC)
+  private static let style: WindowStyle = (base: 0, extended: 0)
+
+  @_Win32WindowText
+  fileprivate var text: String?
```
compnerd, line 15:
```suggestion
  public var text: String?
```

The type itself is `fileprivate` so it is okay for this to be `public`

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
+
+  fileprivate init(frame: Rect) {
```
compnerd, line 17:
```suggestion
  public init(frame: Rect) {
```

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
 
 import WinSDK
 
-public class Label: Control {
+fileprivate class Win32Label: View {
```
compnerd, line 10:
```suggestion
fileprivate class LabelView: View {
```

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
+    super.init(frame: frame, class: Win32Label.class, style: Win32Label.style)
+  }
+}
+
+public class Label: Control {
+  private static let `class`: WindowClass =
+      WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Label")
   private static let style: WindowStyle = (base: WS_TABSTOP, extended: 0)
 
+  private let win32Label: Win32Label
```
compnerd, line 27:
```suggestion
  private let view: LabelView
```

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
 
-  @_Win32WindowText
-  public var text: String?
+  public var text: String? {
+    get { return win32Label.text }
+    set(value) { win32Label.text = value }
+  }
 
   public init(frame: Rect) {
+    win32Label = Win32Label(frame: Rect(origin: .zero, size: frame.size))
```
compnerd, line 46:
```suggestion
    self.view = LabelView(frame: Rect(origin: .zero, size: frame.size))
```

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
     super.init(frame: frame, class: Label.class, style: Label.style)
+    addSubview(win32Label)
```
compnerd, line 48:
```suggestion
    self.addSubview(self.view)
```

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
+    get { return win32Label.font }
+    set(value) { win32Label.font = value }
   }
 
-  @_Win32WindowText
-  public var text: String?
+  public var text: String? {
+    get { return win32Label.text }
+    set(value) { win32Label.text = value }
+  }
```
compnerd, line 43: Are `font`, `text` the only properties we need to forward?

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
+      WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Label")
   private static let style: WindowStyle = (base: WS_TABSTOP, extended: 0)
 
+  private let win32Label: Win32Label
+
+  public override var frame: Rect {
+    didSet {
+      win32Label.frame = Rect(origin: .zero, size: frame.size)
+    }
+  }
```
compnerd, line 33: I think that we should override `subviews` as well to hide the container, WDYT?

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -7,20 +7,45 @@
...
+fileprivate class Win32Label: View {
   private static let `class`: WindowClass = WindowClass(named: WC_STATIC)
+  private static let style: WindowStyle = (base: 0, extended: 0)
+
+  @_Win32WindowText
+  fileprivate var text: String?
+
+  fileprivate init(frame: Rect) {
+    super.init(frame: frame, class: Win32Label.class, style: Win32Label.style)
+  }
```
compnerd, line 19: Does it make sense to make sure that the label has a transparent background?

compnerd (changes requested):

compnerd:
Thinking about this a bit further, is there any advantage to actually deriving the label from `View`?  That is, perhaps it makes more sense to just manually create the `WC_STATIC` window locally and add it as a child and forget about it entirely.  We should be able to just handle the resize as `didSet` observers on the frame.

egorzhdan:
If we get rid of the `LabelView` component, we'll have to reimplement some of `View`'s logic, namely `font` & `text` properties, for the manually created `WC_STATIC` component. Do you think that would still be a better option?

Sources/SwiftWin32/Views and Controls/Label.swift
egorzhdan: I think yes, the rest of them don't seem to affect the internal `WC_STATIC` component.

Sources/SwiftWin32/Views and Controls/Label.swift
egorzhdan: Oh yeah you're right, I haven't thought about that.

Sources/SwiftWin32/Views and Controls/Label.swift
egorzhdan: I guess it does, but `View.backgroundColor` isn't implemented yet, and I think `Color.clear` is mentioned in the doc comments but isn't defined. Perhaps this could be a part of the future change that will bring support for background colors? :)

compnerd:
> If we get rid of the `LabelView` component, we'll have to reimplement some of `View`'s logic, namely `font` & `text` properties, for the manually created `WC_STATIC` component. Do you think that would still be a better option?

True, but, we end up having to do that in senses anyways - we forward the settings to the subview.  At the end of the day, we just need to change the hWnd that the messages are sent to, so that seems no more expensive really, it just is a matter of inlining the call.  If I am not mistaken, instead of forwarding the properties, we will just inline the `SendMessageW` for those cases right?

Sources/SwiftWin32/Views and Controls/Label.swift
compnerd: Sounds good.

Sources/SwiftWin32/Views and Controls/Label.swift
compnerd: The alternative discussed below cleverly sidesteps the issue.

egorzhdan:
> we just need to change the hWnd that the messages are sent to

Hmm, do you mean changing `self.hWnd` to the `WS_STATIC` component's hWnd? I don't think that would work, since the parent component would get overwritten when adding the label as a subview to another view.

> it just is a matter of inlining the call

So basically calling all the WinSDK methods for creating/repositioning the `WS_STATIC` component from `Label.init`?

compnerd:
> Hmm, do you mean changing `self.hWnd` to the `WS_STATIC` component's hWnd? I don't think that would work, since the parent component would get overwritten when adding the label as a subview to another view.
> 
> >

No, that won't work for exactly the reason you mention.  However, rather than having the properties forward to another view, which then just invokes `SendMessageW`, we can inline those into the properties instead and use the stashed HWND for the static label.  You counted a small set of properties which needed to be forwarded, so this should be equivalent.

> So basically calling all the WinSDK methods for creating/repositioning the `WS_STATIC` component from `Label.init`?

No, the thought that I had was actually the insight that we dont care about the settings for the `LabelView`.  The only things we care about are:
- Font
- Foreground Color
- Position

If we set the background color to clear, instead of forwarding the font and foreground color we send the proper message to the emebdded view, and set a property observer on the position, we can entirely ignore the embedded window.

egorzhdan:
Got it, thanks! I think this should now work the way you expect.

compnerd (commented):
I think that this is actually going to be far more important than I first expected.  The `Label` type really should be a composite type which doesn't leak the members.  Thank you so much for working on this!

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -21,21 +21,71 @@ private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSub
...
+    self.staticHWnd = CreateWindowExW(0, WC_STATIC.wide, nil, 0,
+                                      0, 0,
+                                      Int32(staticFrame.size.width),
+                                      Int32(staticFrame.size.height),
+                                      nil, nil, GetModuleHandleW(nil), nil)!
+
+    var staticStyle: LONG = GetWindowLongW(self.staticHWnd, WinSDK.GWL_STYLE)
+    staticStyle |= WS_CHILD
+    staticStyle &= ~LONG(bitPattern: WS_POPUP | WS_CAPTION)
+    _ = SetWindowLongW(self.staticHWnd, WinSDK.GWL_STYLE, staticStyle)
```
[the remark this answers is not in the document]
egorzhdan: Yeah, we actually only need `WS_CHILD` here, so I made it static.

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -21,21 +21,71 @@ private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSub
...
+      // Scale window for DPI
+      var client: Rect = self.frame
+      ScaleClient(rect: &client, for: GetDpiForWindow(self.staticHWnd),
+                  (base: 0, extended: 0))
+
+      // Resize and Position the Window
+      _ = SetWindowPos(self.staticHWnd, nil,
+                       CInt(client.origin.x), CInt(client.origin.y),
+                       CInt(client.size.width), CInt(client.size.height),
+                       UINT(SWP_NOZORDER | SWP_FRAMECHANGED))
```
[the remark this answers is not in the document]
egorzhdan: We can use the parent rect indeed. Fixed it.

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -21,21 +21,71 @@ private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSub
...
     _ = SetWindowSubclass(hWnd, SwiftLabelProc, UINT_PTR(1),
                          unsafeBitCast(self as AnyObject, to: DWORD_PTR.self))
+
+    var staticFrame = frame
+    ClientToWindow(size: &staticFrame.size, for: (base: 0, extended: 0))
+    self.staticHWnd = CreateWindowExW(0, WC_STATIC.wide, nil, 0,
+                                      0, 0,
+                                      Int32(staticFrame.size.width),
+                                      Int32(staticFrame.size.height),
+                                      nil, nil, GetModuleHandleW(nil), nil)!
```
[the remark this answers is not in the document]
egorzhdan: Fixed

compnerd (approved):
Thanks!  I think that this can be simplified a bit further, but we can do that in a follow up clean up.  Assuming it also builds with CMake, I don't think that we need to push this out any further.

compnerd:
Ugh, I think looking into the CMake build can happen later.

PR #316, opened 2020-12-12
UI: expand Menu public interface

egorzhdan:
Store `Identifier` & `Options` as fields of `Menu`. This will be useful for supporting inline submenus.
It also matches the macCatalyst interface.

### DIFF
--- a/Sources/SwiftWin32/UI/Menu.swift
+++ b/Sources/SwiftWin32/UI/Menu.swift
@@ -6,6 +6,7 @@
  **/
 
 import WinSDK
+import class Foundation.NSUUID
 
 extension Menu {
   /// Constants for identifying an application's standard menus.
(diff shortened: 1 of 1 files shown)


Sources/SwiftWin32/UI/Menu.swift
```diff
@@ -246,6 +247,12 @@ extension Menu.Identifier {
   }
 }
 
+extension Menu.Identifier {
+  fileprivate static var generated: Menu.Identifier {
```
compnerd, line 251: Not sure how I feel about this, it may be unexpected the value returned is different each time.  I wonder if it makes sense to have this be a function to make it more obvious.  But, that can be cleaned up in a follow up as this is internal.

compnerd (approved):

Issue #315, opened 2020-12-12
View's `frame` setter has different scaling from `frame` init argument


egorzhdan:
I've configured 200% scaling in Windows settings, and some components in UICatalog now look misplaced:

<img width="283" alt="Screen Shot 2020-12-12 at 19 11 13" src="https://user-images.githubusercontent.com/3801618/101989469-8436e480-3cb1-11eb-9542-8a73907f4f8f.png">

The image view, and the table view cells look unexpectedly. Frame for these components is set via `frame` property, while for the others `frame` initializer argument is used.

Looks like the coordinates & dimensions for these components should have been multiplied by 2.
Perhaps `View.frame.didSet` should perform scaling differently?

compnerd:
The rendering definitely indicates something is wrong.  However, I'm not sure where this would be going wrong.

https://github.com/compnerd/swift-win32/blob/master/Sources/SwiftWin32/UI/View.swift#L113-L125
This indicates that the frame setter does scale the new frame for the DPI

This matches the operations in the initializer:
https://github.com/compnerd/swift-win32/blob/master/Sources/SwiftWin32/UI/View.swift#L186-L193

compnerd:
Playing around with this since I happened to get some scaling when remoting - it seems that its not the fact that we are setting the frame rect via the frame var I think.  It is the resizing that occurs during the reparenting:

https://github.com/compnerd/swift-win32/blob/master/Sources/SwiftWin32/UI/View.swift#L291-L294

The client has already been scaled and sized, but the reparenting will result in the view actually getting resized back.  However, failure to do perform this step fails to correctly size the buttons in Calculator.  Without this operation though, the list view is nearly correct, except the offsets for the rows are incorrect.

egorzhdan:
Ah, that makes sense. Thanks for the explanation!

PR #334, opened 2020-12-18
UI: adjust client area during reparenting

compnerd:
When the view is reparented, we cannot simply query the frame and use
the value.  We must scale the value appropriately for the current DPI
scaling.  This is required since we store the requested scale rather
than the rendered scale (which allows greater flexibility).

Fixes: #315


compnerd:
CC: @egorzhdan

PR #372, opened 2021-01-07
UI: support icons in menu items

egorzhdan:
This PR adds the support for icons displayed in context menu items, on the left side from the item title.

### DIFF
--- a/Sources/SwiftWin32/Menus and Shortcuts/MenuElement.swift
+++ b/Sources/SwiftWin32/Menus and Shortcuts/MenuElement.swift
@@ -73,15 +73,25 @@ public class MenuElement {
 internal class Win32MenuElement {
   private var title: [WCHAR]
   private let submenu: Win32Menu?
+  private let image: BitmapHandle?
 
   internal var info: MENUITEMINFOW
 
-  private init(title: String, submenu: Win32Menu?, fType: Int32) {
+  private init(title: String, image: Image?, submenu: Win32Menu?, fType: Int32) {
     self.title = title.LPCWSTR
     self.submenu = submenu
+
+    let imageHandle: BitmapHandle?
+    if let bitmap = image?.bitmap {
+      imageHandle = BitmapHandle(from: bitmap)
+    } else {
+      imageHandle = nil
+    }
+    self.image = imageHandle
+
     self.info = self.title.withUnsafeMutableBufferPointer {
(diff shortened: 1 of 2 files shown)


Sources/SwiftWin32/Menus and Shortcuts/MenuElement.swift
```diff
@@ -91,25 +101,28 @@ internal class Win32MenuElement {
...
+      submenu = Win32Menu(MenuHandle(owning: CreatePopupMenu()),
+                          items: menu.children)
     } else {
-      self.init(title: element.title, submenu: nil, fType: MFT_STRING)
+      submenu = nil
     }
+    self.init(title: element.title,
+              image: element.image,
+              submenu: submenu,
+              fType: MFT_STRING)
```
compnerd, line 119: Im not sure I understand the point of this refactoring.  More importantly, the actual change is lost in the noise due to the rafactoring.

Sources/SwiftWin32/Views and Controls/ImageView.swift
```diff
@@ -9,8 +9,8 @@ import WinSDK
 import SwiftCOM
 
 extension BitmapHandle {
-  fileprivate convenience init?(from pBitmap: SwiftCOM.IWICBitmapSource?,
-                                hWnd: HWND? = nil) {
+  convenience init?(from pBitmap: SwiftCOM.IWICBitmapSource?,
```
compnerd, line 12: Can we keep this as `internal`?  Is there some other use that you have in mind for this outside of the module?  In the latter, I think that we should at least mark this somehow as being in use externally so it doesn't get dropped.  Perhaps we should consider adding a test case?

Sources/SwiftWin32/Menus and Shortcuts/MenuElement.swift
egorzhdan: It was just to remove code duplication. Reverted it.

Sources/SwiftWin32/Views and Controls/ImageView.swift
egorzhdan: It is already internal, since `BitmapHandle` itself is also internal.
Added an `internal` keyword for better visualization.

compnerd: Oh, right.  Thanks!

compnerd (approved):

PR #376, opened 2021-01-17
Hello World sample: Added a title and a hello world label

yostane:
Hello,
I have added a title and a label to the "hello world" example.
Best regards,

### DIFF
--- a/Examples/HelloWorld/HelloWorld.swift
+++ b/Examples/HelloWorld/HelloWorld.swift
@@ -20,8 +20,10 @@ extension HelloWorld: SceneDelegate {
     let window = Window(windowScene: windowScene)
     self.window = window
 
-    // Add a title
+    // the rootViewController allows add a title and some controls
     window.rootViewController = ViewController()
+
+    // Add a title
     window.rootViewController?.title = "swift-win32 Hello World"
 
     // Add a hello world label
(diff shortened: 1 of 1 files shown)


Examples/HelloWorld/HelloWorld.swift
```diff
@@ -17,7 +17,21 @@ extension HelloWorld: SceneDelegate {
...
-    self.window = Window(windowScene: windowScene)
-    self.window?.makeKeyAndVisible()
+    let window = Window(windowScene: windowScene)
+    self.window = window
+
+    // The rootViewController allows to set a title and add controls
+    window.rootViewController = ViewController()
+
+    // Add a title
+    window.rootViewController?.title = "swift-win32 Hello World"
```
compnerd, line 27: Would you mind using "Swift/Win32 Hello World" instead?
```suggestion
    window.rootViewController?.title = "Swift/Win32 Hello World"
```

Examples/HelloWorld/HelloWorld.swift
```diff
@@ -17,7 +17,21 @@ extension HelloWorld: SceneDelegate {
...
+
+    // Add a hello world label
+    let label = Label(frame: Rect(x: 50.0, y: 100.0, width: 300.0, height: 30.0))
+    label.text = "Hello World with some Swift ❤️"
```
compnerd, line 31: Let's make this purely "Hello World!"  Is there some particular reason to include additional text?  Was there something that you wanted this to demonstrate?
```suggestion
    label.text = "Hello World!"
```

compnerd (changes requested):
I can see some value to adding the title/label, but at the same time, this is intended to just be a skeleton to setup a window.  What exactly is this trying to achieve or what is the motivation here?

Could you please squash the changes into a single commit and add that information into the commit message?

yostane:
Ok, I'll create another PR with the requested information

PR #377, opened 2021-01-18
HelloWorld example: added a title and a label

yostane:
- Main motivation: The "Hello World" example does not show a "hello world" message. This commit fixes this.
- Other motivation: I remember that "hello world" examples in GUI frameworks usually show at least a label.

### DIFF
--- a/Examples/HelloWorld/HelloWorld.swift
+++ b/Examples/HelloWorld/HelloWorld.swift
@@ -9,7 +9,7 @@ import SwiftWin32
 
 @main
 final class HelloWorld: ApplicationDelegate {
-  var window: Window?
+  var window: Window!
 }
 
 extension HelloWorld: SceneDelegate {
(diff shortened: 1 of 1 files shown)


Examples/HelloWorld/HelloWorld.swift
```diff
@@ -17,7 +17,21 @@ extension HelloWorld: SceneDelegate {
              options: Scene.ConnectionOptions) {
     guard let windowScene = scene as? WindowScene else { return }
 
-    self.window = Window(windowScene: windowScene)
-    self.window?.makeKeyAndVisible()
+    let window = Window(windowScene: windowScene)
```
compnerd, line 20: Any reason for the local variable here?  Since this is a class (reference type), it gains nothing but a shadow (and a refcount bump).  Using `self.window` should be equivalent I believe.  Am I missing something?

compnerd (commented):
Thanks for the changes!

Examples/HelloWorld/HelloWorld.swift
yostane: I added the local var to avoid the optional chaining of self.window in the following lines.

without the local var:

```swift
self.window = Window(windowScene: windowScene)
self.window?.rootViewController = ViewController()
self.window?.rootViewController?.title = "swift-win32 Hello World"
self.window?.addSubview(label)
self.window?.makeKeyAndVisible()
```

with the local var

```swift
let window = Window(windowScene: windowScene)
self.window = window
window.rootViewController = ViewController()
window.rootViewController?.title = "swift-win32 Hello World"
window.addSubview(label)
window.makeKeyAndVisible()
```

We can also have a mix of both by making the window property an implicitly unwrapped optionallike this to avoid any chaining:

```swift
@main
final class HelloWorld: ApplicationDelegate {
  var window: Window!
}
   

 self.window.rootViewController = ViewController()
    self.window.rootViewController?.title = "swift-win32 Hello World"
    self.window.addSubview(label)
    self.window.makeKeyAndVisible()

}
```

compnerd: I think that the IUO makes sense to do here, if you think its better to avoid the optional chaining.  I don't seem to be bothered by the optional chaining, but, really, the reason for the window being optional is that it will only be created when the application is setup.

yostane: OK for turning window into an IUO :)

PR #378, opened 2021-01-26
Fix for compnerd/swift-win32/issues/272

yostane:
- Application main loads the delegate name from info.plist
- main: remove delegate param since it's loaded from info.plist
- Calculator: copy info.plist in CMakeLists (calculator exe was not loading before that)

### DIFF
--- a/Sources/SwiftWin32/Application/ApplicationMain.swift
+++ b/Sources/SwiftWin32/Application/ApplicationMain.swift
@@ -43,8 +43,7 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
 @discardableResult
 public func ApplicationMain(_ argc: Int32,
                             _ argv: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>,
-                            _ application: String?,
-                            _ delegate: String?) -> Int32 {
+                            _ application: String?) -> Int32 {
   let hRichEdit: HMODULE? = LoadLibraryW("msftedit.dll".LPCWSTR)
   if hRichEdit == nil {
     log.error("unable to load `msftedit.dll`: \(Error(win32: GetLastError()))")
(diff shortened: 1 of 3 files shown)


Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -43,8 +43,7 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
 @discardableResult
 public func ApplicationMain(_ argc: Int32,
                             _ argv: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>,
-                            _ application: String?,
-                            _ delegate: String?) -> Int32 {
+                            _ application: String?) -> Int32 {
```
compnerd, line 46: I would prefer to retain the ability to explicitly specify the delegate if the user so desires.  I don't think that it really adds much in terms of complexity.  If one is not specified, we can fallback to the `Info.plist` value, and if one is not present, we can default the delegate.

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -58,8 +57,17 @@ public func ApplicationMain(_ argc: Int32,
...
+  if let path = Bundle.main.path(forResource: "Info", ofType: "plist") {
+      if let contents = FileManager.default.contents(atPath: path) {
+        Application.shared.information =
+            try? PropertyListDecoder().decode(Application.Information.self,
+                                              from: contents)
+      }
+  }
+
+  // Setup ApplicationDelegate using the 'SceneDelegateClassName' info from the plist
+  if let delegate = Application.shared.information?.scene?.configurations?[SceneSession.Role.windowApplication.rawValue]?.first?.delegate {
```
compnerd, line 70: The scene delegate and application delegates are separate (or can be).  I don't follow the reasoning for initializing the application delegate from the scene delegate.

compnerd (changes requested):
Thanks for the change!  Would you mind splitting out the CMake change for the `Info.plist` copying to a separate change?

yostane:
> Thanks for the change! Would you mind splitting out the CMake change for the `Info.plist` copying to a separate change?

Ok, I'll close this one and make separate PRs

PR #380, opened 2021-01-28
compnerd/swift-win32/issues/272: Support `PrincipalClass` in `Info.plist` to specify the principal class

yostane:


yostane:
Hello,
As a first step, I added the NSPrincipalClass entry into each example's info.plist. However, I don't know yet what to next do and how to do it 😅 . I need to do some more investigation.

compnerd:
https://github.com/compnerd/swift-win32/blob/main/Sources/SwiftWin32/Application/ApplicationMain.swift#L53-L59 sets up the `Application.shared` instance if one is configured.  If one is not configured, it will fall back to https://github.com/compnerd/swift-win32/blob/main/Sources/SwiftWin32/App%20and%20Environment/Application.swift#L13

yostane:
Thanks, I'll take a look

yostane:
Hello. I tried to merge main into my branch to fix conflicts. However, the conflict still appears. Shall I try a new PR ?

compnerd:
The conflict needs to be resolved one way or another.  I don't see how opening a new PR helps resolve the conflict.   However, it seems like a pretty trivial conflict, which I can help resolve.

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -43,8 +43,7 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
 @discardableResult
 public func ApplicationMain(_ argc: Int32,
                             _ argv: UnsafeMutablePointer<UnsafeMutablePointer<Int8>?>,
-                            _ application: String?,
-                            _ delegate: String?) -> Int32 {
+                            _ application: String?) -> Int32 {
```
compnerd, line 46: I think that leaving the signature as it was is reasonable.  I actually just realized that the signature is not what I intended it to be which is:

```swift
ApplicationMain(_: Int32, _: UnsafeMutablePointer<UnsafeMutablePointer<CChar>?>, _: String?, _: String?) -> Int32
```

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -58,8 +57,17 @@ public func ApplicationMain(_ argc: Int32,
...
+  if let path = Bundle.main.path(forResource: "Info", ofType: "plist") {
+      if let contents = FileManager.default.contents(atPath: path) {
+        Application.shared.information =
+            try? PropertyListDecoder().decode(Application.Information.self,
+                                              from: contents)
+      }
+  }
+
+  // Setup ApplicationDelegate using the 'SceneDelegateClassName' info from the plist
+  if let delegate = Application.shared.information?.scene?.configurations?[SceneSession.Role.windowApplication.rawValue]?.first?.delegate {
```
compnerd, line 70: If a `delegate` is specified, it should take precedence.  I would say that this should be:

```swift
  let AppInfo = Application.shared.information
  let WindowApplicationIndex = SceneSession.Role.windowApplication.rawValue
  if let delegate = delegate ?? AppInfo?.scene?.configurations?[WindowApplicationIndex]?.first?.delegate {
```

compnerd: Note that this is still incorrect - this is setting the *application* delegate to the *scene* delegate.  The two concepts are separate, and should be kept distinct.

yostane: I'm really sorry, I don't know what's happening lol. I did not push this a second time. I'll try again :). Thanks for your patience.

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: I'm really sorry, I don't know what's happening lol. I did not push this a second time. I'll try again :). Thanks for your patience.

PR #383, opened 2021-02-06
build: use new `executableTarget` API

egorzhdan:
It's now possible to explicitly declare executable targets in SwiftPM, starting from Swift 5.4.
This change removes `main.swift` files in the example targets – SwiftPM now treats them as executable even without `main.swift`.


compnerd (approved):
This is great!  This definitely is a very nice cleanup.  As long as the CI is happy, I'm okay with this change.

PR #385, opened 2021-02-14
issues/272: Added support for NSPrincipalClass

### LINKED ISSUE #272: Support `PrincipalClass` in `Info.plist` to specify the principal class
`Info.plist` should be consulted for the principal class name when loading the application.  Currently, we do not do so and assume that the `SwiftWin32.Application` is the principal class.

yostane:
- Fixes #272 
- info.plist is loaded more early in ApplicationMain to allow loading the App class from NSPrincipalClass if avaialble
- Made Application and Responder open to allow for user-defined Application subclass
- Added example app that defines NSPrincipalClass. The example contains also a small readme
- Added an indentation in Information enums as used in swift doc https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html


Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -40,6 +40,25 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
   return WinSDK.WaitMessage()
 }
 
+// Loads info.plist if available into Application.Information
+private func loadInfoPlist() -> Application.Information? {
+  guard let path = Bundle.main.path(forResource: "Info", ofType: "plist") else {
+    log.warning("Failed to find info.plist")
```
compnerd, line 46: I don't think that this really adds much in terms of information - can we add some additional information about the failure?  I think that it should be okay if no `Info.plist` is there by default for the most trivial application.  This would however log a warning.

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -40,6 +40,25 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
...
+    return nil
+  }
+  guard let contents = FileManager.default.contents(atPath: path) else {
+    log.warning("Failed to load contents of info.plist")
```
compnerd, line 50: This warning has nothing actionable.  If we cannot explain why the load failed, I don't think that we should be logging a warning here.

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -40,6 +40,25 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
...
+    return nil
+  }
+
+  return try? PropertyListDecoder().decode(Application.Information.self, from: contents)
+}
+
+// Gets application class name by prioritizing the userPassedApp class name
+func getApplicationClassName(userPassedApp: String?, appInformation: Application.Information?) -> String?{
```
compnerd, line 58: I don't think that the keyword arguments add any value here.  Why the preference for:
`getApplicationClassName(userPassedApp:appInformation:)` rather than `getApplicationClassName(_:_:)`?

Furthermore, the single line function doesn't add any real clarity:

```swift
  let `class`: String = application ?? (Application.shared.Info?.principalClass ?? "Application")
```
seems much clearer - if the user specified one, it takes precedence, otherwise, the Info.plist defined value does.  And should neither be found, "Application" is the default.

Sources/SwiftWin32/Application/Information.swift
```diff
@@ -15,9 +15,9 @@ extension Application {
 
 extension Application.SceneConfiguration: Decodable {
   enum CodingKeys: String, CodingKey {
-  case name = "SceneConfigurationName"
-  case `class` = "SceneClassName"
-  case delegate = "SceneDelegateClassName"
+    case name = "SceneConfigurationName"
+    case `class` = "SceneClassName"
+    case delegate = "SceneDelegateClassName"
```
compnerd, line 20: I don't mind the whitespace change here, but please make that part of a separate commit, that has nothing to do with the handling of the PrincipalClass.

Sources/SwiftWin32/App and Environment/Application.swift
```diff
@@ -6,7 +6,7 @@
  **/
 
 /// The centralised point of control and coordination for running applications.
-public class Application: Responder {
+open class Application: Responder {
```
compnerd, line 9: The change to `open` for this and `Responder` are completely correct, but I don't understand how they are related to handling the `NSPrincipalClass` property.  They should be a separate change IMO.

compnerd (changes requested):
Thanks!

I'm not sure if the example for the custom principal class is particularly valuable.  I don't think that many people would be using that.  If its particularly important, we can always just change the UICatalog demo.

Sources/SwiftWin32/Application/Information.swift
yostane: Indeed :)

Sources/SwiftWin32/App and Environment/Application.swift
yostane: Changing to open is a prerequisite for this MR. If Application and Responder are left "public", the developer cannot subclass "Application", and thus, cannot take advantage of NSPrincipalClass. 

We can make this a separate MR that should be merged before this one. If we accept this MR without this change, the compiler would never allow subclassing Application.

Do you prefer a separate MR ? Both replies are ok for me 😄. If yes, I can create the MR.

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: Do you mean that we should not log at all or log to a different level (debug, info) ?

yostane: The only information that we have when the gaurd fails is that :

> If path specifies a directory, or if some other error occurs, this method returns nil.

("some other error"  XD)

Can I log this for example: "Failed to load info.plist at (path). The path is a directory, or an unknown error occured" ?

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: About adding informations to the log, the apple documentation says about the return value:

> or nil if the file could not be located.

Can I log this for example ? `Failed to locate optional Info.plist`

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: Nice, I'll do the modifications 👍

yostane:
You're welcome 😄 

I created the example mainly for my testing. So, it also ok for me to remove it for now add it later (to UICatalog or in another way) if necessary. 
However, if we do that, will the developer know of this possibility and how to to do it ? I would suggest at least to have some kind of documentation. What do you think ?

Sources/SwiftWin32/Application/ApplicationMain.swift
compnerd: I don't think that this should be a warning, if you want to log it at a "lower" log level, that is fine.

Sources/SwiftWin32/Application/ApplicationMain.swift
compnerd: How do you determine if it is a directory based on that?  I would say that if you want, you can try to see if FileManager will provide the additional details.  Otherwise, I don't think that this should be logged as a warning.

Sources/SwiftWin32/App and Environment/Application.swift
compnerd: Sure, a separate PR for that is perfectly fine.

compnerd:
Documentation overall is something that would be great to have, but we don't have that currently.  I think that a separate PR to add documentation is a great idea, and I would definitely be excited to see that.

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: We can get additional details with : [fileExists(atPath:isDirectory:)](https://developer.apple.com/documentation/foundation/filemanager/1410277-fileexists)

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: On a second though, I think that we should not use yet `Application.shared.Info?.principalClass` because we are currently trying to load `Application.shared`. That's why I passed the `appInformation` argument.

yostane:
Before starting writing a documentation, do you have preferences or recommendations about the documentation format / tools and the documentation structure / plan ? In that case, I think that it should be better if you create an issue where you specify those elements. What do you think ?

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: Loading the app class from :  `Application` because classes are prefixed with app names I guess.

```log
SwiftWin32/ApplicationMain.swift:83: Fatal error: unable to find application class: Application
```

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: After some tests, I have found that it if the info.plst is a folder, `Bundle.main.path` fails. So I will just for now reduce the log to debug.

compnerd: That seems expensive - you are doing a file system operation to check if it is a directory which doesn't give you any additional information and you should immediately spot.  This is about constructing a standard bundle as defined by Foundation, this is not something specific and special to this framework.  This simply adds clutter and adds unnecessary diagnostic code and log spew.

Sources/SwiftWin32/Application/ApplicationMain.swift
compnerd: The changes change the order in which the information is loaded.  If there is a local variable, that is sufficient as a replacement ;-).

```swift
let `class`: String = application ?? (info?.principalClass ?? "Application")
```

Examples/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
 add_subdirectory(Calculator)
 add_subdirectory(HelloWorld)
 add_subdirectory(UICatalog)
-
```
compnerd, line 4: Unnecessary change :)

Sources/SwiftWin32/App and Environment/Application.swift
compnerd: Rebasing should get rid of this change

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -40,6 +40,20 @@ private func WaitMessage(_ dwMilliseconds: UINT) -> Bool {
   return WinSDK.WaitMessage()
 }
 
+// Loads info.plist if available into Application.Information
+private func loadInfoPlist() -> Application.Information? {
+  guard let path = Bundle.main.path(forResource: "Info", ofType: "plist") else {
+    log.debug("unable to locate info.plist")
+    return nil
+  }
+  guard let contents = FileManager.default.contents(atPath: path) else {
```
compnerd, line 49: I think that unless you have concrete reasons why the file failed to deserialize I think that this is better:

```swift
  var information: Application.Information?
  if let path = Bundle.main.path(forResource: "Info", ofType: "plist") {
    if let contents = FileManager.default.contents(atPath: path) {
      information = try? PropertyListDecoder().decode(Application.Information.self, from: contents)
    }
  }
```

At that point, I don't think that this is sufficiently simple to just inline into the single call site.

Sources/SwiftWin32/Application/ApplicationMain.swift
```diff
@@ -50,14 +64,22 @@ public func ApplicationMain(_ argc: Int32,
     log.error("unable to load `msftedit.dll`: \(Error(win32: GetLastError()))")
   }
 
-  // Setup Application
-  if let application = application {
+  let information = loadInfoPlist()
+
+  // Setup Custom Application class if available by giving priority to user passed app.
+  // Otherwise, use NSPrincipal class specified in 'info.plist'. Otherwise, use Application
+  if let application = application != nil ? application : information?.principalClass {
```
compnerd, line 71: The `nil`-cascading operation is clearer:

```swift
let application: String = application ?? (information?.principalClass ?? NSStringFromClass(Application.self))
guard let instance = NSClassFromString(application) else {
  log.error("unable to find application class \(application)")
  exit(EXIT_FAILURE)
}
Application.shared = (instance as! Application.Type).init()
Application.shared.information = information
```

compnerd (changes requested):

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: Indded I agree

Sources/SwiftWin32/Application/ApplicationMain.swift
yostane: Nice ! I'll take a look

Examples/CMakeLists.txt
yostane: Ideed XD

compnerd (approved):
Unfortunate that it took so many rounds to convey the simplicity of the change, but thank you for the perseverance and the change :)

yostane:
You're welcome. It's ok for me to take the necessary time, it's not like there are project deadlines to meet XD.

Thanks for the kind and interesting reviews.

PR #390, opened 2021-02-22
Documentation: add some documentation about bundle layouts

compnerd:
Add some documentation describing the bundle layout format which is both
expected by CoreFoundation and thus by Foundation as well as the
Swift/Win32 framework.  This is a minimal description which should
hopefully make it easier for others to contribute.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.swift text eol=lf
+*.md text eol=lf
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @yostane

yostane:
Nice :). Sorry, I didn't get the notification.

PR #393, opened 2021-02-25
Added PrincipalClass documentation

yostane:


Documentation/PrincipalClass.md
```diff
@@ -0,0 +1,44 @@
...
+    let message = "This is a message from MyApplication subclass"
+}
+```
+
+- If you use CMake, do not forget to register `MyApplication.swift` in CMakeLists.txt.
+
+```txt
+add_executable(...
+... MyApplication.swift)
+```
```
compnerd, line 22: I think that this belongs in a separate section if you want to cover how to use this with CMake.

Documentation/PrincipalClass.md
```diff
@@ -0,0 +1,44 @@
...
+
+- In `Info.plist`, define a `PrincipalClass` entry with a String value of `{AppName}.{CustomClassName}`. For example, if the application is named `CustomPrincipalClass`, then we can register `MyApplication` (defined above) as follows:
```
compnerd, line 24: I don't think that `{AppName}.{ClassName}` is the right pattern.  It is supposed to be `{Module}.{Class}` I believe.  There is no guarantee that the application and module name are the same.

Documentation/PrincipalClass.md
```diff
@@ -0,0 +1,44 @@
...
+<plist version="1.0">
+  <dict>
+    <key>PrincipalClass</key>
+    <string>CustomPrincipalClass.MyApplication</string>
+    <!-- other content -->
+  </dict>
+</plist>
+```
+
+- You can reference your custom application subclass anywhere in you app by casting `Application.shared` to your custom application class. For example, here is how to print the `message` property of the `MyApplication` subclass defined above:
```
compnerd, line 38: This is documenting `Application` rather than the `PrincipalClass`.  I think that should be separate.

compnerd (changes requested):

Documentation/PrincipalClass.md
yostane: Maybe something like this shows how to get a reference to the PrincipalClass without documenting Application   ?

`The application now uses the class defined in PrincipalClass instread of the default one. For example, here is how to print the `message` property of the `MyApplication` subclass defined above:`

Documentation/PrincipalClass.md
yostane: Yes I agree :)

Documentation/PrincipalClass.md
compnerd: I think that it would be more useful to document why/when it is useful to actually provide a custom principal class.  This would allow you to focus on the details of the principal class rather than how to get the instance.

yostane: From my research, I found that PrincipalClass is mostly used in plugin scenarios for macOS / Cocoa's loadable bundles or in extensions. The princial class allows to get the entry point of the plugin.
Is it possible to create a plugin or extension on SwiftWin 32 ?

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingCode/Tasks/CreatingBundles.html

compnerd:
Thanks!  This looks great.  I think that a good follow up here would be to actually describe what the principal class is, but this looks good to merge as is, we can enhance the documentation further in subsequent changes.

compnerd (approved):

compnerd:
Oh, another item that could be useful to add is how to control this via the API rather than `Info.plist` (although, for most cases, the plist approach is better).

PR #441, opened 2021-03-31
Views and Controls: dispatch actions for Label

rjpilgrim:
I believe that this is all that is needed to implement the `STN_CLICKED` event.

Example usage:

```
let label=  Label(frame: Rect(x: 0, y: 0, width: 1000, height: 32))
label.font = Font(name: "Consolas", size: Font.systemFontSize)
label.text = "Hello"
label.addTarget(self, action: pressMe(_:), for: .primaryActionTriggered)
window.addSubview(label)


func pressMe(_: Label?) {
    MessageBoxW(nil, "Swift/Win32 Demo!".wide,
                "Swift/Win32 MessageBox!".wide, UINT(MB_OK))
  }
```

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/Label.swift
+++ b/Sources/SwiftWin32/Views and Controls/Label.swift
@@ -7,9 +7,22 @@
 
 import WinSDK
 
+private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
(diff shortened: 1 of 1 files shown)


compnerd:
Interesting, does WC_STATIC not have a `WM_LBUTTONUP` event without `SS_NOTIFY`?  That only generates a `STN_CLICKED` for the parent I thought.

compnerd:
Cleaning up the styles can be done in a follow up I think.  Thanks for the patch!

PR #444, opened 2021-04-06
reloadData() deletes cells currently in TableView and replaces them with new ones from TableViewDataSource

rjpilgrim:
reloadData() in TableView currently does nothing except pad the TableView subviews with the new cells.

I needed to add a handle for WM_DeleteItem to the TableView proxy to handle deleting the old cells. This message is triggered by the LB_RESETCONTENT message at the top of reload data.

Test scenario in UICatalog:

```
@main
final class UICatalog: ApplicationDelegate, SceneDelegate {
...
  var funStrings : [String] = ["Iggy", "David", "Lou"]
  var funEnabled : [Bool] = [false, false, false]

  
  lazy var funToggles : () -> [(Button) -> () -> Void] =
    {
      [unowned self] in
      var myList : [(Button) -> () -> Void] = []
      for i in 0..<funEnabled.count {
        myList.append(
        {
          _ in
        { 
          self.funEnabled[i] = !self.funEnabled[i]
          self.tableview.reloadData()
        }
        }
        )
      }
      return myList
    }
}

extension UICatalog: TableViewDataSource {

  public func tableView(_ tableView: TableView,
                        numberOfRowsInSection section: Int) -> Int {
    return 3
  }

  public func tableView(_ tableView: TableView,
                        cellForRowAt indexPath: IndexPath) -> TableViewCell {
    let cell = TableViewCell(style: .default, reuseIdentifier: nil)
    let button = Button(frame: Rect(x: 0, y: 0, width: 80, height: 32),
                           title: "\(funEnabled[indexPath.row] ? funStrings[indexPath.row] : "Button") \(indexPath.row)")
    button.addTarget(button, action: self.funToggles()[indexPath.row],
                          for: .primaryActionTriggered)

    cell.addSubview(button)
    return cell
  }
}
```

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
+++ b/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
@@ -31,6 +31,7 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         let rctRect: RECT = lpDrawItem.pointee.rcItem
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
+        // TODO(rjpilgrim) figure out why this set to isHidden is needed on second call to reloadData()
         view.isHidden = false
         return DefWindowProcW(view.hWnd, UINT(WM_PAINT), 0, 0)
       }
--- a/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
+++ b/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
@@ -31,6 +31,7 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         let rctRect: RECT = lpDrawItem.pointee.rcItem
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
+        view.isHidden = false
         return DefWindowProcW(view.hWnd, UINT(WM_PAINT), 0, 0)
       }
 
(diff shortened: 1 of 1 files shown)


Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
```diff
@@ -31,6 +31,7 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         let rctRect: RECT = lpDrawItem.pointee.rcItem
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
+        view.isHidden = false
```
compnerd, line 34: Hmm, not sure I understand why the `isHidden` needs to be set here.  Could you add a comment explaining it?

compnerd (approved):
Thanks!  I'd like to understand the need for the adjusting of the `isHidden` property, but functionality-wise this looks great.

compnerd:
@rjpilgrim would you mind rebasing the change please?  Once it has been rebased, we should be able to merge it

Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
rjpilgrim: I am not positive why it is needed, but the window does not display on the second load without it (the first load works fine).

rjpilgrim:
I just performed the rebase

Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
rjpilgrim: I did a run test where a button outside of the table triggered the reload data. With this test, all deinitializers (and therefore DestroyWindow()) were called for every TableCells and their children before any new TableCell was created or parented into the TableView, and yet the new Views are still not visible without `view.isHidden = false`. I had thought maybe lingering Views could be affecting behavior, but this does not seem to be the csae.

compnerd: Hmm, would you mind putting in a TODO comment to investigate and remove the visibility (unless you are implying that this isn't needed, in which case we can just remove it).

rjpilgrim: I added the comment. It is needed currently, and I will continue to look into why

compnerd (approved):
Thank you, this looks great!

PR #445, opened 2021-04-07
Views and Controls: simplify `Label` construction

compnerd:
Further simplify the construction of the `Label` type.  We use the
window constructor to set the style and the parent on the instead of
applying them manually after the construction.  We can also avoid
computing the size of the window and instead use the size of the parent
window itself.  In theory, we could go further - the public API does not
permit the constructing the window parented, so we can rely on the
re-parenting to trigger the resizing, allowing the property observer to
trigger the resizing.


compnerd:
CC: @egorzhdan

PR #448, opened 2021-04-08
Only call isHidden when needed

rjpilgrim:
I have found it the reason the TableCells were hidden after reloadData() was called a second time. For the first call which occurs when the data source is initially assigned to the TableView, a call to `window.makeKeyAndVisible()`  follows. `makeKeyAndVisible()` will then go through and make the TableView and its populated cells visible. On the second call to reloadData(), the call to `makeKeyAndVisible()` does not occur, and so it is necessary to manually set the Cells to Visible.

I have also added support for decoding the uMsgs into their Win32 macro names. This makes it far easier to understand the messages coming into a view at runtime.

Example usage for decoding messages:

```
private let SwiftButtonProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
  let button: Button? = unsafeBitCast(dwRefData, to: AnyObject.self) as? Button
  print("Button receive message: \(getStringforMsg(uMsg))")
  ...
}
```


compnerd:
Thanks for investigating this @rjpilgrim.  Would you mind splitting this up into two separate PRs please?

rjpilgrim:
Okay, I will go ahead and do that.

Issue #447, opened 2021-04-07
Wire up `Stepper` to `TableView` in UICatalog


compnerd:
We can make the stepper control the number of entries in the TableView for now.  This would allow for testing the `reloadData` in `TableView` more easily.  This could be then used to exercise the path identified as being incorrect in #444.

compnerd:
CC: @rjpilgrim

PR #449, opened 2021-04-08
UICatalog: the stepper controls the number of rows of the table

yostane:
Fixes: #447 
Requires: #452

### DIFF
--- a/Examples/UICatalog/UICatalog.swift
+++ b/Examples/UICatalog/UICatalog.swift
@@ -169,13 +169,14 @@ Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deseru
 
   private func stepperValueDidChange(_ stepper: Stepper) {
     self.stepperLabel.text = String(Int(stepper.value))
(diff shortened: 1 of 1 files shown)


compnerd:
Very cool @yostane!

compnerd:
The change looks good to me overall.  What do you think of initializing the stepper to 2 or 3 so that there are elements in the TableView by default?

compnerd (approved):

yostane:
It's weird, `self.stepper.value = 2.0` does not seem to work. It looks like the `SendMessageW` call does not change the POS.

compnerd:
Ugh, because the parameters are swapped :-(: https://docs.microsoft.com/en-us/windows/win32/controls/udm-setpos32

Is a trivial fix, just swap the two parameters, would you like to put up a PR?

yostane:
Okay I'll do it :). Thanks for the help :)

compnerd (approved):
Thank you!

yostane:
You're welcome :)

Issue #442, opened 2021-03-31
Views and Controls: `Label` should derive from `View`


compnerd:
`Label` should be a `View`, not a `Control`.  Event handling can be done through `GestureRecognizer`s.

PR #450, opened 2021-04-08
Label inherits View and clean event handling code

yostane:
Fixes: #442

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/Label.swift
+++ b/Sources/SwiftWin32/Views and Controls/Label.swift
@@ -7,20 +7,7 @@
 
 import WinSDK
 
-private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
-  let label: Label? = unsafeBitCast(dwRefData, to: AnyObject.self) as? Label
-
-  switch uMsg {
-  case UINT(WM_LBUTTONUP):
-    label?.sendActions(for: .primaryActionTriggered)
-  default:
-    break
-  }
-
-  return DefSubclassProc(hWnd, uMsg, wParam, lParam)
-}
-
-public class Label: Control {
+public class Label: View {
   private static let `class`: WindowClass =
       WindowClass(hInst: GetModuleHandleW(nil), name: "Swift.Label")
   private static let style: WindowStyle = (base: WS_TABSTOP, extended: 0)
@@ -59,8 +46,6 @@ public class Label: Control {
 
   public init(frame: Rect) {
     super.init(frame: frame, class: Label.class, style: Label.style)
-    _ = SetWindowSubclass(hWnd, SwiftLabelProc, UINT_PTR(1),
-                         unsafeBitCast(self as AnyObject, to: DWORD_PTR.self))
 
     let size = self.frame.size
     self.hWnd_ = CreateWindowExW(0, WC_STATIC.wide, nil, DWORD(WS_CHILD),


compnerd:
Thanks for the patch @yostane!  Does this not break the current mechanism for handling events?  I think that the long term solution here is to use gesture recognizers and attach one to the button, but for the mean time, it would be nice to have something that continues to function.

yostane:
Hi you're welcome :).

Ah ok I understand. What should be done is: 
- Make Label inherit from View
- Keep the current event handling mechanism (because GestureRecognizer is not implemented yet)

If that's what is expected, I will update my patch if possible.

compnerd:
Yeah, I think that would be the best middle ground where we dont have the desired state, but keep the ability to detect the click.  I think that follow ups for adding GestureRecognizers would be wonderful :)

yostane:
Hi @compnerd . Can you take a look at my last modifications please ?

I transferred some event handling code from Control into Label to allow the developer to use the `addTarget` method on the Label. I have added a click handler on the the Label of the HelloWorld example and it responds to the click.
Is this how it should be done ? It that's ok as a temporary solution, I will remove some test code as well as the the WIP status

Thanks,

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -85,7 +73,130 @@ public class Label: Control {
...
+      }
+    }
+  }
+
+  public func callAsFunction(sender: Label, event: Control.Event) {
+    self.method(instance)(sender as! Source, event)
+  }
+}
+
+private let SwiftLabelProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
```
compnerd, line 167: Why did you move this below the class?

Sources/SwiftWin32/Views and Controls/Label.swift
```diff
@@ -59,13 +46,14 @@ public class Label: Control {
...
     super.init(frame: frame, class: Label.class, style: Label.style)
-    _ = SetWindowSubclass(hWnd, SwiftLabelProc, UINT_PTR(1),
-                         unsafeBitCast(self as AnyObject, to: DWORD_PTR.self))
 
     let size = self.frame.size
     self.hWnd_ = CreateWindowExW(0, WC_STATIC.wide, nil, DWORD(WS_CHILD),
                                  0, 0, CInt(size.width), CInt(size.height),
                                  self.hWnd, nil, GetModuleHandleW(nil), nil)!
+
+    _ = SetWindowSubclass(hWnd, SwiftLabelProc, UINT_PTR(1),
```
compnerd, line 55: Was there a reason for this move?

compnerd (commented):
This does add a lot of code to `Label` - and it seems most of it can be extracted into an extension.  What do you think of pulling it all out into an extension and putting it into a separate file (say `Label+Temporary.swift`?  I do have some of the work for the gesture recognizer, we should be able to start considering adding the `TapGestureRecognizer` in place instead.

Sources/SwiftWin32/Views and Controls/Label.swift
yostane: Oups, I put it back in the wrong location 😁

Sources/SwiftWin32/Views and Controls/Label.swift
yostane: Oups, I put it back in the wrong location 😁

yostane:
I tried moving code in Label+Temporary but this seems more complex than expected because of visibility problems. The current Laabel+Temporary file contains code that did require changes.

Instead of spending more time on this temporary solution, maybe it would be better to postpone this MR and focus more on gesture recognizer. What do you think ?

compnerd:
Postponing and focusing on gesture recognizer sounds like a good plan to me.  @egorzhdan  had previously looked at the gesture recognizer bits for menus, not sure if he has any thoughts about what exactly would be the minimum work needed to get it working.

egorzhdan:
Yeah, I've looked into handling menu item clicks before, but I don't have a working implementation of it.
IIUC for menus that would require handling `WM_COMMAND` events & deducing the menu item that was clicked based on `LOWORD(wParam)`, but for labels the event could be different.

PR #453, opened 2021-04-08
Only set isHidden when needed

rjpilgrim:
I have found the reason the TableCells were hidden after reloadData() was called a second time. For the first call which occurs when the data source is initially assigned to the TableView, a call to window.makeKeyAndVisible() follows. makeKeyAndVisible() will then go through and make the TableView and its populated cells visible. On the second call to reloadData(), the call to makeKeyAndVisible() does not occur, and so it is necessary to manually set the Cells to Visible.

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
+++ b/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
@@ -32,8 +32,8 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
         // Setting isHidden is necessary for TableCells generated after initial call to Window.makeKeyAndVisible()
-        let parenthWnd = GetParent(view.hWnd)
-        if IsWindowVisible(parenthWnd) && !IsWindowVisible(view.hWnd) {
+        let hWndParent = GetParent(view.hWnd)
+        if IsWindowVisible(hWndParent) && !IsWindowVisible(view.hWnd) {
             view.isHidden = false
         }
         return DefWindowProcW(view.hWnd, UINT(WM_PAINT), 0, 0)
--- a/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
+++ b/Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
@@ -31,7 +31,7 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         let rctRect: RECT = lpDrawItem.pointee.rcItem
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
-        //setting isHidden is necessary for TableCells generated after initial call to Window.makeKeyAndVisible()
+        // Setting isHidden is necessary for TableCells generated after initial call to Window.makeKeyAndVisible()
         let parenthWnd = GetParent(view.hWnd)
         if IsWindowVisible(parenthWnd) && !IsWindowVisible(view.hWnd) {
             view.isHidden = false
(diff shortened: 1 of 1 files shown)


Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
```diff
@@ -31,8 +31,11 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
         let rctRect: RECT = lpDrawItem.pointee.rcItem
         _ = SetWindowPos(view.hWnd, nil, rctRect.left, rctRect.top, 0, 0,
                          UINT(SWP_NOSIZE))
-        // TODO(rjpilgrim) figure out why this set to isHidden is needed on second call to reloadData()
-        view.isHidden = false
+        //setting isHidden is necessary for TableCells generated after initial call to Window.makeKeyAndVisible()
```
compnerd, line 34: Could you please add a space after the `//` and use sentence capitalization please?

Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
```diff
@@ -31,8 +31,11 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
...
+        let parenthWnd = GetParent(view.hWnd)
```
compnerd, line 35:
```suggestion
        let hWndParent = GetParent(view.hWnd)
```

Sources/SwiftWin32/Views and Controls/Table Views/TableView.swift
```diff
@@ -31,8 +31,11 @@ private let SwiftTableViewProxyWindowProc: WNDPROC = { (hWnd, uMsg, wParam, lPar
...
+        if IsWindowVisible(parenthWnd) && !IsWindowVisible(view.hWnd) {
```
compnerd, line 36:
```suggestion
        if IsWindowVisible(hWndParent) && !IsWindowVisible(view.hWnd) {
```

compnerd (changes requested):

compnerd:
I have found that there is an issue with lifetimes that was introduced with `WM_DELETEITEM` handling.  It is clear that we need to handle that message to avoid memory leaks and correct the rendering, but it appears that the cell is being destroyed before being reparented.  Some tracing shows that at the time of unparenting the cell we get: 
view: SwiftWin32.TableViewCell, hWnd: Optional(0x0000000000040670), IsWindow: false
and that results in `SetParent` failing.  Would you mind investigating that further subsequently?

compnerd:
I think that I understand what this change is doing.  I think that this is a reasonable approach, though it is indicating that there is a small problem to the lifecycle of the `TableViewCell`.  When is the right time to make the cell visible?  Should this be a generic thing that occurs in `View` when the view is parented?

compnerd (approved):

rjpilgrim:
When you say "the cell is destroyed before being reparented" do you mean that the new cell coming is destroyed before `SetParent()` is called in `addSubview(cell)`? Or are you referring to an issue in the call to `view.removeFromSuperview()` in the WM_DELETEITEM handler?

> I have found that there is an issue with lifetimes that was introduced with `WM_DELETEITEM` handling. It is clear that we need to handle that message to avoid memory leaks and correct the rendering, but it appears that the cell is being destroyed before being reparented. Some tracing shows that at the time of unparenting the cell we get:
> view: SwiftWin32.TableViewCell, hWnd: Optional(0x0000000000040670), IsWindow: false
> and that results in `SetParent` failing. Would you mind investigating that further subsequently?

compnerd:
> When you say "the cell is destroyed before being reparented" do you mean that the new cell coming is destroyed before `SetParent()` is called in `addSubview(cell)`? Or are you referring to an issue in the call to `view.removeFromSuperview()` in the WM_DELETEITEM handler?
> 
> > I have found that there is an issue with lifetimes that was introduced with `WM_DELETEITEM` handling. It is clear that we need to handle that message to avoid memory leaks and correct the rendering, but it appears that the cell is being destroyed before being reparented. Some tracing shows that at the time of unparenting the cell we get:
> > view: SwiftWin32.TableViewCell, hWnd: Optional(0x0000000000040670), IsWindow: false
> > and that results in `SetParent` failing. Would you mind investigating that further subsequently?

The problem occurs in the `removeFromSuperview` which is invoked in the `VM_DELETEITEM`.

rjpilgrim:
> The problem occurs in the `removeFromSuperview` which is invoked in the `VM_DELETEITEM`. I suspect that the Window has been destroyed but we somehow have a reference to it?

Maybe also WM_DELETEITEM is being called more than once for some reason for a given View. 

If so,
```
if view.superview != nil {
  view.removeFromSuperview()
}
```
would fix that issue. 

I will try to see if this is the case.

PR #454, opened 2021-04-08
Add decoder for giving name for messages sent to WindowProcs

rjpilgrim:
I have added support for decoding the uMsgs into their Win32 macro names. This makes it far easier to understand the messages coming into a view at runtime.

Example usage for decoding messages:

```
private let SwiftButtonProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
  let button: Button? = unsafeBitCast(dwRefData, to: AnyObject.self) as? Button
  print("Button receive message: \(getStringforMsg(uMsg))")
  ...
}
```


Sources/SwiftWin32/Support/Win32MessageDecoder.swift
```diff
@@ -0,0 +1,1030 @@
+/**
+ * SPDX-License-Identifier: BSD-3-Clause
+ **/
```
compnerd, line 3: Please copy the standard header in the project, replacing the author as yourself.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
```diff
@@ -0,0 +1,1030 @@
...
+import WinSDK
+
+class Win32MessageDecoder {
```
compnerd, line 6: Why is this a `class`?  There doesn't seem to be anything that needs a reference type here.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
```diff
@@ -0,0 +1,1030 @@
...
+	static func getStringforMsg(_ uint: UINT) -> String {
```
compnerd, line 7: I wonder if we can create a `WindowMessage` type, and then use `CustomStringConvertible` instead?  WDYT?

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
```diff
@@ -0,0 +1,1030 @@
...
+		if let message = messages[Int(uint)] {
+			return message
+		}
+		return "Unknown Message"
+	}
+
+	static let messages : [Int : String] = {
```
compnerd, line 14: The style of the project uses 2-space indents, could you reformat the file please?  I can try to add a swift format configuration file and try to write up some style documentation later.

compnerd (changes requested):

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
rjpilgrim: This has been done.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
rjpilgrim: I do not see much value in creating a new type here, considering that the only use of the type would be for the use of its text representation. Such a type could be useful if the callbacks themselves could be made to work with WindowMessage, but I expect you would have to make a C wrapper to handle such functionality, which may be more effort than it is worth.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
rjpilgrim: I restyled the file with spaces.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
compnerd: I just don't see a way to constrain the conformance.  Adding the conformance to `UINT` is overly broad.  I suppose we could do `WindowMessage.debugDescription(for:)` instead.

Sources/SwiftWin32/Support/Win32MessageDecoder.swift
```diff
@@ -0,0 +1,1033 @@
...
+
+struct Win32MessageDecoder {
+  static func getStringforMsg(_ uint: UINT) -> String {
+    if let message = messages[Int(uint)] {
+      return message
+    }
+    return "Unknown Message"
+  }
+
+  static let messages : [Int : String] = {
```
compnerd, line 18: What does this accomplish exactly?  Im not sure I understand this part still.  Can we not just use the static mapping below?

rjpilgrim: The static mapping has duplicates, I am not sure the best way to handle these. I am currently pruning the Static Mapping of _FIRST and _LAST entries, as those are especially unhelpful, but I am not sure if there is a way to resolve messages to their name considering I have only access to a number for the sake of decoding.

rjpilgrim: The static mapping is transposed from the following source: https://wiki.winehq.org/List_Of_Windows_Messages

rjpilgrim:
I have went ahead and renamed the Decoder to WindowMessage, with a function debugDescription. This is the implementation I believe you were dsecribing.

Usage now becomes:

```
private let SwiftButtonProc: SUBCLASSPROC = { (hWnd, uMsg, wParam, lParam, uIdSubclass, dwRefData) in
  let button: Button? = unsafeBitCast(dwRefData, to: AnyObject.self) as? Button
  print("Button receive message: \(WindowMessage.debugDescription(for: uMsg))")
  ...
}
```

compnerd:
Thank you, that is indeed the interface that I had in mind.  Would you mind rebasing the change so that it is easier to review?

rjpilgrim:
I have rebased on top of the current main.

rjpilgrim:
I don't think that I've used the best method for resolving the file name change during the rebase. The commit history is confusing.

rjpilgrim:
I've shrunk the commits down into one.

Sources/SwiftWin32/Support/WindowMessage.swift
```diff
@@ -0,0 +1,1025 @@
...
+ **/
+
+import WinSDK
+
+struct WindowMessage {
+  static func debugDescription(for uint: UINT) -> String {
+    if let message = messages[Int(uint)] {
+      return message
+    }
+    return "Unknown Message"
```
compnerd, line 15: This could be written more succinctly as:

```suggestion
    return messages[Int(uMsg)] ?? "Unknown Message"
```

Sources/SwiftWin32/Support/WindowMessage.swift
```diff
@@ -0,0 +1,1025 @@
...
+  }
+
+  static let messages : [Int : String] = {
+    var messages : [Int : String] = [:]
```
compnerd, line 19:
```suggestion
    var messages: [Int:String] = [:]
```

Sources/SwiftWin32/Support/WindowMessage.swift
```diff
@@ -0,0 +1,1025 @@
...
+    var keys = Set<Int>()
+    for (key, value) in keyValues {
+      if !keys.contains(key) {
+        keys.insert(key)
+        messages[key] = value
+      }
+    }
+    return messages
```
compnerd, line 27: I'm still not convinced of the value of this.  The duplicated values end up discarded anyways, why have them listed in the first place?

compnerd, line 27: This seems more complicated than necessary.

```suggestion
    return Dictionary<Int, String>(keyValues, uniquingKeysWith: { (first, _) in first })
```

Sources/SwiftWin32/Support/WindowMessage.swift
```diff
@@ -0,0 +1,1025 @@
...
+    8500  :  "OCM_CTLCOLORLISTBOX",
+    8501  :  "OCM_CTLCOLORBTN",
+    8502  :  "OCM_CTLCOLORDLG",
+    8503  :  "OCM_CTLCOLORSCROLLBAR",
+    8504  :  "OCM_CTLCOLORSTATIC",
+    8720  :  "OCM_PARENTNOTIFY",
+    32768 :  "WM_APP",
+    52429 :  "WM_RASDIALEVENT"
+  ]
+}
```
compnerd, line 1025: Missing newline.

```suggestion
}
```

Sources/SwiftWin32/Support/WindowMessage.swift
```diff
@@ -0,0 +1,1025 @@
...
+ * Copyright © 2021 Ryan Allan <rjallan@g.ucla.edu>
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ **/
+
+import WinSDK
+
+struct WindowMessage {
+  static func debugDescription(for uint: UINT) -> String {
```
compnerd, line 11:
```suggestion
  static func debugDescription(for uMsg: UINT) -> String {
```

compnerd (changes requested):

rjpilgrim:
I am currently working to remove all duplicates from Dictionary literal, allowing it to be used to instantiate messages. Duplicated messages will show up as follows: "MSG_FTS_JUMP_QWORD || RB_SHOWBAND || TB_GETTOOLTIPS" for instance for 1059.

rjpilgrim:
Reworked the approach to duplicate messages. Some of the resulting strings are a bit long, but I think it is important to preserve the information.

compnerd (approved):
Sure, this works for now.  We can always clean it up subsequently.  I think that there might actually be some use in explicitly building up some infrastructure for decoding messages.

Issue #425, opened 2021-03-21
Views and Controls: Implement PickerView


compnerd:
Implement a `PickerView` view which enables users to select a value from a set.

PR #456, opened 2021-04-09
PickerView: add an initial implementation of PickerView

compnerd:
This adds a constrained implementation of PickerView.  It sketches out
the various pieces that are needed to get a working picker view.
Ideally, we would use the `CBS_OWNERDRAWVARIABLE` to render the views
individually sized, but that seems to cause an issue with the draw
operation, where the draw is never invoked for each item.  This
currently restricts us to equal sized views rather than variable sized
views.

The other important missing piece here is the lack of support for
multi-component pickers.  This should be possible to do in practice
since the backing widget tree is a ListBox and an Edit container.  We
should be able to render more complicated views accordingly.

~~An item that this work has uncovered is the lack of respect for
`sizeThatFits` involving a label.~~

Resolves: #425

### DIFF
--- a/Examples/UICatalog/UICatalog.swift
+++ b/Examples/UICatalog/UICatalog.swift
@@ -60,6 +60,9 @@ final class UICatalog: ApplicationDelegate, SceneDelegate {
       TableView(frame: Rect(x: 4.0, y: 310.0, width: 254.0, height: 48.0),
                 style: .plain)
 
+  lazy var pickerview: PickerView =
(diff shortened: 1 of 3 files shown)


compnerd:
Current results with a few workarounds in place:
image

compnerd:
At this point, I think that the missing piece is rendering the selected view in the edit location.  This is trickier since we need to re-parent the window to render the content or figure out how to render the window in two spots.

compnerd:
The piece that is left here is rendering the selected item.

PR #463, opened 2021-04-14
Gestures: add GestureRecognizer and GestureRecognizerDelegate

compnerd:
This adds the two base types for the hierarchy to support implementing
gesture recognizers.  This is meant to provide the proper interaction
detection for Label components.  Currently, Label incorrectly derives
from Control, when it should derive from View.

This works towards a solution for #442.


compnerd:
CC: @yostane

PR #474, opened 2021-04-27
Add hitTest() functions to View

rjpilgrim:
Found a need for this in working with TableView and its TableViewCells.

Since the TableViewCell if we follow UIKit's example is only supposed to have a single "ContentView", embedding a row of Buttons into a single TableViewCell will require adding them as subviews of that parent ContentView. The Mouse handler for the parent view, however, in my testing was blocking access to the underlying buttons in this case.

This additional case to the default View message handler will allow button messages to be correctly routed to the underlying child subview. I perform some arithmetic in order to adjust the x and y coordinates passed to the correct context of the subview.

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -72,6 +72,30 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle) 
   rect = Rect(from: r)
 }
 
+private func UndoTransform(view: View, for point: Point) -> Point {
+  let myDeterminant = view.transform.a * view.transform.d - view.transform.b * view.transform.c
+  let (newA, newB, newC, newD) = (view.transform.d/myDeterminant, -view.transform.b/myDeterminant, -view.transform.c/myDeterminant, view.transform.a/myDeterminant)
+  let (newTX, newTY) = (-newA * view.transform.tx - newC * view.transform.ty, -newB * view.transform.tx - newD * view.transform.ty)
+  let myTransformInverted = 
+    AffineTransform( a: newA,  b: newB,
+                     c: newC,  d: newD,
+                    tx: newTX, ty: newTY)
+
+  return point.applying(myTransformInverted)
+}
+
+private func UndoTransform(view: View, for rect: Rect) -> Rect {
+  let myDeterminant = view.transform.a * view.transform.d - view.transform.b * view.transform.c
+  let (newA, newB, newC, newD) = (view.transform.d/myDeterminant, -view.transform.b/myDeterminant, -view.transform.c/myDeterminant, view.transform.a/myDeterminant)
+  let (newTX, newTY) = (-newA * view.transform.tx - newC * view.transform.ty, -newB * view.transform.tx - newD * view.transform.ty)
+  let myTransformInverted = 
+    AffineTransform( a: newA,  b: newB,
+                     c: newC,  d: newD,
+                    tx: newTX, ty: newTY)
+
+  return rect.applying(myTransformInverted)
+}
+
 extension View {
   /// Options to specify how a view adjusts its content when its size changes.
   public enum ContentMode: Int {
@@ -332,6 +356,13 @@ public class View: Responder {
     set { self.frame = Rect(origin: Point(x: self.frame.origin.x - newValue.x,
                                           y: self.frame.origin.y - newValue.y),
                             size: self.frame.size) }
+  }
+
+  /// Specifies the transform applied to the view, relative to the center of its bounds
+  public var transform: AffineTransform = AffineTransform.identity {
+    didSet {
+      fatalError("\(#function) not yet implemented")
+    }
   }
 
   // MARK - Managing the View Hierarchy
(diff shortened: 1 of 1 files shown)


compnerd:
Hmm, I'm afraid I don't follow what is going on here.  The `Button` type will handle the dispatch of the `WM_LBUTTONUP` with the `addTarget(_:action:for:)` on the `.touchUpInside`.  I think that a test case for this is probably useful.  I believe that it should be possible to use XCTest to add a test case.

That said, the thing that seems to be doing is hit testing, which is not yet implemented.  However, that implementation should not be based in `Button` but in `View`, where we should be able to map the even to the proper view.

rjpilgrim:
My claim about this affecting the nested buttons was incorrect, as you pointed out. `Button`s seem to be working with or without my change. This change does allow `Label`s to work, however.

I agree that using XCTest functionality with hit testing would be helpful for proving this sort of behavior out. For now, here is a reworking of UICatalog that proves the behavior:

```
@main
final class UICatalog: ApplicationDelegate, SceneDelegate {
  var window: Window!
  ...
  private func pressLabel(_: Label) {
    MessageBoxW(nil, "Swift/Win32 Demo!".wide,
                "Swift/Win32 MessageBox!".wide, UINT(MB_OK))
  }
}

extension UICatalog: TableViewDataSource {
  public func tableView(_ tableView: TableView,
                        numberOfRowsInSection section: Int) -> Int {
    return Int(stepper.value)
  }

  public func tableView(_ tableView: TableView,
                        cellForRowAt indexPath: IndexPath) -> TableViewCell {
    let cell = TableViewCell(style: .default, reuseIdentifier: nil)
    let contentView = View(frame: Rect(x: 0, y: 0, width: 80, height: 32))

    let button = Button(frame: Rect(x: 0, y: 0, width: 40, height: 32), title: "\(indexPath.row)A")
    button.addTarget(self, action: UICatalog.pressMe(_:), for: .primaryActionTriggered)

    let labelB =  Label(frame: Rect(x: 40, y: 0, width: 40, height: 32))
    labelB.text = "\(indexPath.row)B"
    labelB.addTarget(self, action: UICatalog.pressLabel(_:), for: .primaryActionTriggered)

    contentView.addSubview(button)
    contentView.addSubview(labelB)
    
    cell.addSubview(contentView)
    return cell
  }
}
```
Without the change, the Label does not respond to being clicked on.

compnerd:
Ah, for multiple labels, this makes sense.  I think that the Label handling is not what it should be currently.  It really should be handled by means of a tap gesture recognizer is my understanding.  Is there any documentation to indicate otherwise?

I'm finishing up some of the type definition for the TapGestureRecognizer type.  Once that is complete, wiring that up to `Label` sounds like the better approach.  The work that you are doing here I believe is in the wrong location.  It seems that the hit testing should be generalized and moved into `View`.  That should also expose the information to `Label` and we can wire that through the tap gesture recognizer.

https://developer.apple.com/documentation/uikit/uiview/1622469-hittest may be of interest for a review.

rjpilgrim:
I agree. It is best to rework the label to the new approach before worrying how it is handled in this case. 

You are right that my change could be easily reworked to implement the hitTest function on View, which I will go ahead and do.

compnerd:
Thinking a bit more about this, it seems like it is a combination of `hitTest(_:with:)`, `convert(_:to:)`, `convert(_:from:)`.

rjpilgrim:
Update: `hitTest()` was rather easy to implement using the `ChildWindowFromPoint` call. Your last comment inspired me to investigate the `convert()` from functions from `UIView` as well...

With the current `View` implementation, all that is needed for these functions is the translation arithmetic in my original commit. If we keep consistency with `UIView`, however, then we must look at the `transform` property to account for a potential rotation of the coordinate systems. 

The `transform` property is not currently implemented in `View`, and to do so would probably require using the Windows GDI `SetWorldTransform()` function. For now, I propose adding `transform` as a private constant to View initialized to the identify matrix (this is an `AfflineTransform` that does nothing to the coordinate system). In this way, a convert implementation can be done referencing the transform which will also work when the transform property is implemented.

compnerd:
Why could we not just add the `transform` property, initialize it to `.identity` and add a `didSet` observer that `fatalError`s?  I have a few such stubs in various places.

rjpilgrim:
Using the didSet observer there is a good idea.

I have implemented the convert function, but am seeing unexpected behavior after testing that seems to be a result of some translation occurring during drawing. 

I decided to try a test of converting the `textfield` bounds Rect from UICatalog (even though bounds is not implemented this should be exactly `Rect(x: 0.0, y: 0.0, width: 254.0, height: 17.0)`) to the coordinate system of `password`. The line of code I ran is:
```
print("Here is convert: \(textfield.convert(Rect(x: 0.0, y: 0.0, width: 254.0, height: 17.0), to: password))")
```

Since these two views are identical except 21 units apart, I expect this to return `Rect(x:0.0, y:-21.0, width: 254.0, 17.0)`. I replicated these two element in UIKit on an IPad and this is exactly what it returns. In my function's case, though, it returns `Rect(x:0.0, y:-26.0, width: 254.0, 17.0)`.

In my implementation of `convert`, I am using `GetWindowRect()` to find the screen coordinates of the two Views and then use this information to translate the view. The difference in y returned by `GetWindowRect()` for the two views is 26, meaning one or both of the views have been moved away from the position specified by the `frame` property.

If this is expected behavior, I can either keep the function as is, or I could avoid `GetWindowRect()` entirely, travel up the root parent for both passed Views, and accumulate the translation amount from the frame property. This would make the function behavior line up with UIKit, and Screen Coordinate nuances would remain hidden from the user.

rjpilgrim:
I believe the frame accumulation approach is the way to go, but I have pushed for reference.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,30 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
...
+  let myDeterminant = view.transform.a * view.transform.d - view.transform.b * view.transform.c
+  let (newA, newB, newC, newD) = (view.transform.d/myDeterminant, -view.transform.b/myDeterminant, -view.transform.c/myDeterminant, view.transform.a/myDeterminant)
+  let (newTX, newTY) = (-newA * view.transform.tx - newC * view.transform.ty, -newB * view.transform.tx - newD * view.transform.ty)
+  let myTransformInverted = 
+    AffineTransform( a: newA,  b: newB,
+                     c: newC,  d: newD,
+                    tx: newTX, ty: newTY)
+
+  return point.applying(myTransformInverted)
+}
```
compnerd, line 85: This looks like normal operations for matricies.  A number of these should be standalone functions in https://github.com/compnerd/swift-win32/blob/main/Sources/SwiftWin32/CG/AffineTransform.swift.  Things like `AffineTransformInvert` (or as an extension method on `AffineTransform` named `invert`).

I really need to get around to writing that style guide.  I try to keep the column width to 80 whenever possible, but I do sometimes exceed it if there is no good way to break the expression (e.g. it would be in middle of a word).  Please indent continued lines 2 (that is 4 spaces, not 2).

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -334,6 +358,13 @@ public class View: Responder {
                             size: self.frame.size) }
   }
 
+  /// Specifies the transform applied to the view, relative to the center of its bounds
```
compnerd, line 361: Nit: missing `.` at the end of the doc comment.

compnerd (commented):
I think that we can split this up into two separate changes: one to add the conversion routines and one to add the hit testing.  That should allow us to focus on the specifics.  We could even split out the math operations into its own standalone change if you feel like that.

As to the difference, I think that we should try to understand why there is that difference.  If we need to, then walking the hierarchy and computing the value that way is valid, but if we can avoid that, it is preferable.

rjpilgrim:
Update: convert() was implemented in separate pull request. hitTest() is probably best implemented manually on top of an existing `Rect.contains()` method, because after the transforms are applied `ChildWindowFromPointEx` will no longer work.

compnerd:
@rjpilgrim are you planning on resuming this PR?

compnerd:
Given that there are now conflicts and there has been no progress on this for a while, I'm going to close this.  However, I would still be happy to have an updated version of this change merged when ready.

PR #479, opened 2021-05-01
Add inverted() to AffineTransform

rjpilgrim:
Per your suggestion, I have decided to split the commit for hitTest() and convert() into different parts. I realized that CGAffineTransform from CoreGraphics includes an inverted function. So it makes sense to add it to AffineTransform as well.

### DIFF
--- a/Sources/SwiftWin32/CG/AffineTransform.swift
+++ b/Sources/SwiftWin32/CG/AffineTransform.swift
@@ -59,6 +59,18 @@ public struct AffineTransform {
                            tx: self.tx * transform.a + self.ty * transform.c + transform.tx,
                            ty: self.tx * transform.b + self.ty * transform.d + transform.ty)
   }
+
+  public func inverted() -> AffineTransform {
+    let determinant = a * d - b * c
+    let (newA, newB, newC, newD) = (d/determinant, -b/determinant,
+                                   -c/determinant,  a/determinant)
+    let (newTX, newTY) = (-newA * tx - newC * ty, -newB * tx - newD * ty)
+    let invertedTransform = 
+      AffineTransform( a: newA,  b: newB,
+                       c: newC,  d: newD,
+                      tx: newTX, ty: newTY)
+    return invertedTransform
+  }
 }
 
 extension AffineTransform: Equatable {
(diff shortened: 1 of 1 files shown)


Sources/SwiftWin32/CG/AffineTransform.swift
```diff
@@ -59,6 +59,18 @@ public struct AffineTransform {
                            tx: self.tx * transform.a + self.ty * transform.c + transform.tx,
                            ty: self.tx * transform.b + self.ty * transform.d + transform.ty)
   }
+
+  public func inverted() -> AffineTransform {
+    let determinant = a * d - b * c
+    let (newA, newB, newC, newD) = (d/determinant, -b/determinant,
+                                   -c/determinant,  a/determinant)
+    let (newTX, newTY) = (-newA * tx - newC * ty, -newB * tx - newD * ty)
```
compnerd, line 67: I'm not sure I understand the reason for the tuples here.  Why not just split these into individual statements?  I realize that they are related, but whitespace could delimit that.

Would you mind just moving this function into an extension to `AffineTransform`?

Sources/SwiftWin32/CG/AffineTransform.swift
```diff
@@ -59,6 +59,18 @@ public struct AffineTransform {
...
+    let invertedTransform = 
+      AffineTransform( a: newA,  b: newB,
+                       c: newC,  d: newD,
+                      tx: newTX, ty: newTY)
+    return invertedTransform
```
compnerd, line 72:
```suggestion
    let determinant = self.a * self.d - self.b * self.c
  
    let a = self.d / determinant
    let b = -self.b / determinant
    let c = -self.c / determinant
    let d = self.a / determinant
  
    return AffineTransform(a: a, b: b, c: c, d: d,
                           tx: -a * self.tx * c * self.ty,
                           ty: -b * self.tx - d * self.ty)
    return invertedTransform
```

Seems a bit more concise and avoids an unnecessary set of tuples to be created.  The spacing also makes it easier to follow.

compnerd:
I think that with the suggestion (or if there is something that I am misunderstanding about why the tuples are needed or the let is needed), this would be okay to merge.  It would be nice to follow this up with some tests.

compnerd:
@rjpilgrim sorry about taking over this change, but I'd like to add some additional test coverage.  I don't want to have you deal with the fallout of the changes that follow.  I'm going to apply the suggestions and merge this change.  If something is wrong, feel free to send a follow up PR to repair the damage I may have caused, or if there is something which deserves further discussion, we can continue to discuss the items.

rjpilgrim:
No worries

PR #484, opened 2021-05-01
Add convert() functions to View

rjpilgrim:
I have reimplemented the convert functions with the new `inverted()` function using the frame accumulation approach I described. Hopefully my use of a tuple is not controversial. I think it makes more sense this time (not sure why I like tuples).

I will also add test cases for these functions later tonight to the Tests folder. They returned the same numbers that the `UIView` functions do in my tests on UICatalog.

### DIFF
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -786,7 +786,8 @@ public class View: Responder {
 
   // MARK - Converting Between View Coordinate Systems
 
-  /// Converts a point from the receiver's coordinate system to that of the specified view.
+  /// Converts a point from the receiver’s coordinate system to that of the
+  /// specified view.
   public func convert(_ point: Point, to view: View?) -> Point {
     guard view != nil || self.window != nil else { return point }
 
@@ -841,12 +842,14 @@ public class View: Responder {
                 .applying(WindowBasedTransform(for: view).inverted())
   }
 
-  /// Converts a point from the coordinate system of a given view to that of the receiver.
+  /// Converts a point from the coordinate system of a given view to that of the
+  /// receiver.
   public func convert(_ point: Point, from view: View?) -> Point {
     return view?.convert(point, to: self) ?? self.window?.convert(point, to: self) ?? point
   }
 
-  /// Converts a rectangle from the receiver’s coordinate system to that of another view.
+  /// Converts a rectangle from the receiver’s coordinate system to that of
+  /// another view.
   public func convert(_ rect: Rect, to view: View?) -> Rect {
     guard view != nil || self.window != nil else { return rect }
 
@@ -905,11 +908,11 @@ public class View: Responder {
                .applying(WindowBasedTransform(for: view).inverted())
   }
 
-  /// Converts a rectangle from the coordinate system of another view to that of the receiver.
+  /// Converts a rectangle from the coordinate system of another view to that of
+  /// the receiver.
   public func convert(_ rect: Rect, from view: View?) -> Rect {
     return view?.convert(rect, to: self) ?? self.window?.convert(rect, to: self) ?? rect
   }
-
 
   // MARK - Responder Chain
 
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -759,6 +759,91 @@ public class View: Responder {
         self.subviews.lazy.compactMap { $0.viewWithTag(tag) }.first
   }
 
+  // MARK - Converting Between View Coordinate Systems
+
+  /// Converts a point from the receiver’s coordinate system to that of the
+  /// specified view.
+  public func convert(_ point: Point, to view: View?) -> Point {
+    if let view = view {
+      // If the view is itself, then the point is already in the correct
+      // coordinate system.
+      if view == self { return point }
+
+      // In the case of an immediate relation to the view, just account for the
+      // frame offset.
+      if let superview = self.superview, superview == view {
+        // `p - self.bounds.origin` creates a vector from the top-left of
+        // `self`.  Apply the current transform to the vector, and then return
+        // the point relative to the top-left of `view`.
+
+        // +--- view ---+
+        // | +- self -+ |
+        // | |   p    | |
+        // | +--------+ |
+        // +------------+
+
+        // `bounds` is in the coordinate space of self and `frame` is in the
+        // coordinate space of the superview.  In this case, the superview is
+        // `view`, which is the destination coordinate space.  We simply map
+        // the transformed point p from the coordinate space of `self` into
+        // the destination coordinate space of `view` using `self.frame`.
+        return (point - self.bounds.origin).applying(self.transform) + self.frame.origin
+      } else if let superview = view.superview, superview == self {
+        // `p - view.frame.origin` creates a vector from the top-left of `self`.
+        // Invert any transformations of `self` to the vector, and then return
+        // the point relative to `view.`
+
+        // +--- self ---+
+        // | +- view -+ |
+        // | |   p    | |
+        // | +--------+ |
+        // +------------+
+
+        // `frame` is in the coordinate space of the superview.  Because `self`
+        // is the superview of `view`, `view.frame` is the coordinate space of
+        // `self`.  We locate `point` in the coordinate space of `self`, undo
+        // any local transformation, and then relocate it in the coordinate
+        // space of the destination view.
+        return (point - view.frame.origin).applying(self.transform.inverted()) + view.bounds.origin
+      }
+    }
+
+    func WindowBasedOrigin(for view: View?) -> Point {
+      var origin: Point = .zero
+      guard var view = view else { return origin }
+      while let superview = view.superview {
+        // Map the view's bounds to the superview and compute the vector for the
+        // location.  This must be transformed by the transformation matrix for
+        // the superview.
+        origin = origin
+               + (view.convert(view.bounds.origin, to: superview) - superview.bounds.origin)
+                    .applying(superview.transform)
+        view = superview
+      }
+      return origin
+    }
+
+    return point - (WindowBasedOrigin(for: view) - WindowBasedOrigin(for: self))
+  }
+
+  /// Converts a point from the coordinate system of a given view to that of the
+  /// receiver.
+  public func convert(_ point: Point, from view: View?) -> Point {
+    fatalError("\(#function) not yet implemented")
+  }
+
+  /// Converts a rectangle from the receiver’s coordinate system to that of
+  /// another view.
+  public func convert(_ rect: Rect, to view: View?) -> Rect {
+    fatalError("\(#function) not yet implemented")
+  }
+
+  /// Converts a rectangle from the coordinate system of another view to that of
+  /// the receiver.
+  public func convert(_ rect: Rect, from view: View?) -> Rect {
+    fatalError("\(#function) not yet implemented")
+  }
+
   // MARK - Responder Chain
 
   public override var next: Responder? {
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -828,18 +828,19 @@ public class View: Responder {
             + view.bounds.center
       }
     }
-    
 
     return point.applying(WindowBasedTransform(for: self))
                 .applying(WindowBasedTransform(for: view).inverted())
   }
 
+  /// Converts a point from the coordinate system of a given view to that of the receiver.
   public func convert(_ point: Point, from view: View?) -> Point {
     return view?.convert(point, to: self) ?? self.window?.convert(point, to: self) ?? point
   }
-  
+
+  /// Converts a rectangle from the receiver’s coordinate system to that of another view.
   public func convert(_ rect: Rect, to view: View?) -> Rect {
-    guard view != nil || self.window != nil { return point }
+    guard view != nil || self.window != nil else { return rect }
 
     if let view = view {
       // If the view is itself, then the point is already in the correct
@@ -892,14 +893,11 @@ public class View: Responder {
       }
     }
 
-    if view == nil && self.window == nil {
-      return rect
-    }
-
     return rect.applying(WindowBasedTransform(for: self))
                .applying(WindowBasedTransform(for: view).inverted())
   }
-  
+
+  /// Converts a rectangle from the coordinate system of another view to that of the receiver.
   public func convert(_ rect: Rect, from view: View?) -> Rect {
     return view?.convert(rect, to: self) ?? self.window?.convert(rect, to: self) ?? rect
   }
--- a/Sources/SwiftWin32/Views and Controls/View.swift
+++ b/Sources/SwiftWin32/Views and Controls/View.swift
@@ -72,6 +72,24 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle) 
   rect = Rect(from: r)
 }
 
+private func WindowBasedTransform(for view: View?) -> AffineTransform {
+  guard var view = view else {return AffineTransform.identity}
+  var transform = AffineTransform.identity
+  while let superview = view.superview {
+    // Create a single transform that places this view in the coordinate
+    // space of the view furthest up in its tree.
+    transform = transform
+        .concatenating(AffineTransform(translationX: -view.bounds.center.x,
+                                       y: -view.bounds.center.y))
+    transform = transform.concatenating(view.transform)
+    transform = transform
+        .concatenating(AffineTransform(translationX: view.center.x,
+                                      y: view.center.y))
+    view = superview
+  }
+  return transform
+}
+
 extension View {
   /// Options to specify how a view adjusts its content when its size changes.
   public enum ContentMode: Int {
@@ -352,9 +370,9 @@ public class View: Responder {
 
   /// The bounds rectangle, which describes the view’s location and size in its
   /// own coordinate system.
-  public var bounds: Rect {
+  public var bounds: Rect /*{
     didSet { fatalError("\(#function) not yet implemented") }
-  }
+  }*/
 
   /// The center point of the view's frame rectangle
   public var center: Point {
@@ -366,9 +384,9 @@ public class View: Responder {
 
   /// Specifies the transform applied to the view, relative to the center of its
   /// bounds.
-  public var transform: AffineTransform = .identity {
+  public var transform: AffineTransform = .identity /*{
     didSet { fatalError("\(#function) not yet implemented") }
-  }
+  }*/
 
   // MARK - Managing the View Hierarchy
 
@@ -761,8 +779,7 @@ public class View: Responder {
 
   // MARK - Converting Between View Coordinate Systems
 
-  /// Converts a point from the receiver’s coordinate system to that of the
-  /// specified view.
+  /// Converts a point from the receiver's coordinate system to that of the specified view.
   public func convert(_ point: Point, to view: View?) -> Point {
     if let view = view {
       // If the view is itself, then the point is already in the correct
@@ -770,11 +787,11 @@ public class View: Responder {
       if view == self { return point }
 
       // In the case of an immediate relation to the view, just account for the
-      // frame offset.
+      // center offset and transform.
       if let superview = self.superview, superview == view {
-        // `p - self.bounds.origin` creates a vector from the top-left of
-        // `self`.  Apply the current transform to the vector, and then return
-        // the point relative to the top-left of `view`.
+        // `p - self.bounds.center` undos any translation done by the bounds of
+        // `self`. Apply the current transform to the shifted point, and 
+        // then return the point relative to `self.center`.
 
         // +--- view ---+
         // | +- self -+ |
@@ -782,16 +799,18 @@ public class View: Responder {
         // | +--------+ |
         // +------------+
 
-        // `bounds` is in the coordinate space of self and `frame` is in the
+        // `bounds` is in the coordinate space of self and `center` is in the
         // coordinate space of the superview.  In this case, the superview is
         // `view`, which is the destination coordinate space.  We simply map
         // the transformed point p from the coordinate space of `self` into
-        // the destination coordinate space of `view` using `self.frame`.
-        return (point - self.bounds.origin).applying(self.transform) + self.frame.origin
+        // the destination coordinate space of `view` using `self.center`.
+        return (point - self.bounds.center).applying(self.transform)
+            + self.center
       } else if let superview = view.superview, superview == self {
-        // `p - view.frame.origin` creates a vector from the top-left of `self`.
-        // Invert any transformations of `self` to the vector, and then return
-        // the point relative to `view.`
+        // `p - view.center` shifts the point relative to the center of `view` 
+        // as seen by `self`. Invert any transformations performed by `view` 
+        // on the point, and then return the point relative to the center of 
+        // `view.bounds.`
 
         // +--- self ---+
         // | +- view -+ |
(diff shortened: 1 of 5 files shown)


Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,20 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
   rect = Rect(from: r)
 }
 
+private func GetGlobalOffset(_ view: View) -> (Double, Double) {
+  var x = view.frame.origin.x
+  var y = view.frame.origin.y
+  var currentView : View? = view.superview
```
compnerd, line 78:
```suggestion
  var currentView: View? = view.superview
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,20 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
...
+  while (currentView != nil) {
```
compnerd, line 79:
```suggestion
  while currentView != nil {
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,20 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
...
+    if let view = currentView {
+      x = x + view.frame.origin.x
+      y = y + view.frame.origin.y
+      currentView = view.superview
+    }
```
compnerd, line 84:
```suggestion
    x = x + view?.frame.origin.x ?? 0
    y = y + view?.frame.origin.y ?? 0
    currentView = view?.superview ?? view
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,20 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
...
+  }
+  return (x, y)
```
compnerd, line 86: This feels like it could be pretty concisely written using a reduce over a lazy sequence.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +744,112 @@ public class View: Responder {
...
+  // MARK - Converting Between View Coordinate Systems
+
+  /// Converts a point from the receiver's coordinate system to that of the specified view.
+  public func convert(_ point: Point, to view: View?) -> Point {
+    var newPoint = point.applying(self.transform.inverted())
+
+    let (myX, myY) = GetGlobalOffset(self)
+
+    var xTranslate = myX
+    var yTranslate = myY
```
compnerd, line 756:
```suggestion
    var (xTranslate, yTranslate) = GetGlobalOffset(self)
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +744,112 @@ public class View: Responder {
...
+    }
+
+    newPoint.x = newPoint.x + xTranslate
+    newPoint.y = newPoint.y + yTranslate
+
+    if let convertView = view {
+        newPoint = newPoint.applying(convertView.transform)
+    }
+
+    return newPoint
```
compnerd, line 772:
```suggestion
    return view == nil ? newPoint : newPoint.applying(convertView.transform)
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -72,6 +72,20 @@ private func ScaleClient(rect: inout Rect, for dpi: UINT, _ style: WindowStyle)
   rect = Rect(from: r)
 }
 
+private func GetGlobalOffset(_ view: View) -> (Double, Double) {
```
compnerd, line 75: Should we use `Point` here?  That would simplify the handling of this point across the rest of the file.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +744,112 @@ public class View: Responder {
...
+    return newPoint
+  }
+
+  /// Converts a point from the coordinate system of a given view to that of the receiver.
+  public func convert(_ point: Point, from view: View?) -> Point {
+    var newPoint = point
+
+    if let convertView = view {
+      newPoint = point.applying(convertView.transform.inverted())
+    }
```
compnerd, line 781:
```suggestion
    var newPoint = point.applying(view?.transform.inverted() ?? .identity)
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +744,112 @@ public class View: Responder {
...
+  }
+
+  /// Converts a rectangle from the receiver's coordinate system to that of another view.
+  public func convert(_ rect: Rect, to view: View?) -> Rect {
+    var newRect : Rect = rect.applying(self.transform.inverted())
+
+    let (myX, myY) = GetGlobalOffset(self)
+
+    var xTranslate = myX
+    var yTranslate = myY
```
compnerd, line 808:
```suggestion
    var (xTranslate, yTranslate) = GetGlobalOffset(self)
```

compnerd (commented):
I think a bit of exposition (comments) in the changes here would be useful.  Especially currently.  The mapping seems very complicated.

rjpilgrim:
I have reimplemented the functions. To avoid the tuple, a Vector made the most sense to me, because we are creating a translation component rather than an actual point. I also added some operators for Vector in order to make the code more concise.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +742,76 @@ public class View: Responder {
...
+
+    if let convertView = view {
+      // Subract the called view's global offset.
+      // The result is the needed translation from us to them.
+      offset = offset - GetGlobalOffset(convertView)
+    }
+
+    newPoint = newPoint + offset
+
+    return newPoint.applying(view?.transform ?? .identity)
```
compnerd, line 761: Hmm, thinking a bit further about this, could we not just do something like:

```swift
var point: POINT = POINT(from: point)
_ = withUnsafeMutablePointer(to: &point) {
  MapWindowPoints(self.hWnd, view?.hWnd, $0.baseAddress, UINT($0.count)
}
return Point(from: point)
```
The same pattern can be used for the other variants, they are just changing the first two parameters to `MapWindowPoints`.

rjpilgrim: > Hmm, thinking a bit further about this, could we not just do something like:
> 
> ```swift
> var point: POINT = POINT(from: point)
> _ = withUnsafeMutablePointer(to: &point) {
>   MapWindowPoints(self.hWnd, view?.hWnd, $0.baseAddress, UINT($0.count)
> }
> return Point(from: point)
> ```
> 
> The same pattern can be used for the other variants, they are just changing the first two parameters to `MapWindowPoints`.

I can test this implementation, but I am worried that this will return the strange behavior I found before, where the results do not line up with those evident in the frame values.

rjpilgrim: As I feared, the usage of `MapWindowPoints()` causes things to be scaled away from values in the frame property. There is an incongruity between the frame and screen coordinates that currently causes this. Until this is fixed, my `GetGlobalOffset` does not actually return screen coordinates.

compnerd: Sure, if its incorrect, we can stick with the approach that you have.

rjpilgrim:
I have moved the operators to new files. I could not come up with a better name that I liked for `GetGlobalOffset()`. `GetTopLevelFrameOffset()` was the best I could think of but it seems verbose.

compnerd:
Oh, I'm sorry about not being clear.  I just meant a different extension group in the same file.  I guess a separate file is fine too.

Sources/SwiftWin32/Views and Controls/View.swift
compnerd: I guess that is what I'm left trying to puzzle out.  What exactly does the resulting value represent.  I think that this mapping is complex enough to warrant some set of test cases so that when we do subsequently change this, we know that we have it correct.

rjpilgrim: Yes I have a file ViewTest that I am working on adding to the Tests folder. Is there a better way to organize tests for `View`? I can work on adding tests tomorrow.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +742,76 @@ public class View: Responder {
...
+
+    if let convertView = view {
+      // Add the called view's global offset.
+      // The result is the needed translation from them to us.
+      offset = offset + GetGlobalOffset(convertView)
+    }
+
+    newRect.origin = newRect.origin + offset
+
+    return newRect.applying(self.transform)
```
compnerd, line 812:
```suggestion
  return view?.convert(rect, to: self) ?? rect
```

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -723,6 +742,76 @@ public class View: Responder {
...
+
+    if let convertView = view {
+      // Add the called view's global offset.
+      // The result is the needed translation from them to us.
+      offset = offset + GetGlobalOffset(convertView)
+    }
+
+    newPoint = newPoint + offset
+
+    return newPoint.applying(self.transform)
```
compnerd, line 778:
```suggestion
  return view?.convert(point, to: self) ?? point
```

compnerd (commented):
I think that we might be able to get away with 2 implementations with your approach.  If we can figure out the last bit of difference in the mapping computation, with `MapWindowPoints`, we should be able to get away with 1 (point is just the rect conversion, getting one of the points).  Am I mistaken about this?

Sources/SwiftWin32/Views and Controls/View.swift
compnerd: I don't have an idea of how to best handle the test organization yet as there aren't very many.  I think that adding a new module is fine for now, we can reorganize the tests later.

Sources/SwiftWin32/Views and Controls/View.swift
rjpilgrim: Unfortunately, the one implementation approach will be difficult to handle the nil case. From the UIView.convert() documentation: "If view is nil, this method instead converts to window base coordinates."

compnerd: The same applies to `convert(_:to:)`.  Seems like I missed that bit, however, that is the cascading nil value.  The cascading nil case should therefore convert it to the window's coordinate system.  I think that re-reading that, I'm almost certain that we really want to have the single implementation (for point and rect) because these really are just inversions.

rjpilgrim:
Update: I have run into issues attaching the UnitTest to a SceneDelegate to run the tests. I am doing some research to find a view to perform tests on `View` which will be essential. I have pasted the test that I attempted and its error output at the bottom of the post. I don't fully understand the cause of the crash, but it seems to be crashing in the `TextField` initializer during the call to `CreateWindowExW()`. I believe if the test is running on an actual running Windows appliation in some way these tests will be resolved. 

Regarding the single implementation: when `convert(rect:myRectfrom:nil)` is called on a view, that view needs to go ahead and calculate its GlobalOffset (as I've defined it) and subtract it from the origin of the passed rect. There is no equivalent `convert(to:)` that will perform this particular operation, because `convert(to:)` always assumes to be adding the GlobalOffset to the view on which it is called. At best, you could have a guard case at the top of `convert(from:)` checking for a nil `from` where the else case will do this GlobalOffset subtraction and then return.

On a final point, my testing of UIKit behavior where Windows are not shared seems to be returning undefined behavior. As such, I cannot really reverse engineer the function in that case. Instead, I may return a Rectangle or Point set to all zeros.

My View test case (I tried two different approaches to initializing the Window and neither work):

```
  func testConvertSameParent() {
    /*let session = SceneSession(identifer: "test",
        role: SceneSession.role.windowApplication)
    let windowScene = WindowScene(session: session, 
        connectionOptions: Scene.ConnectionOptions())*/

    //var window = Window(windowScene: windowScene)
    var window = Window(frame: Rect(x:0.0, y:0.0, width:1000, height: 1000))

    var textfield: TextField =
      TextField(frame: Rect(x: 4.0, y: 92.0, width: 254.0, height: 17.0))

    var password: TextField =
      TextField(frame: Rect(x: 4.0, y: 113.0, width: 254.0, height: 17.0))

    textfield.text = "Test"
    password.text = "Test"

     

    window.addSubview(password)
    window.addSubview(textfield)

    let boundsRect = Rect(x: 0, y: 0, width: 254.0, height: 17.0)

    var convertRect = textfield.convert(boundsRect, to: password)
    var expectedConvertRect = Rect(x:0.0, y:-21.0, width: 254.0, height: 17.0)

    XCTAssertEqual(convertRect, expectedConvertRect)
  }
```
My error output:
```
Test Suite 'ViewTests' started at 2021-05-02 20:48:19.231
Test Case 'ViewTests.testConvertSameParent' started at 2021-05-02 20:48:19.231
SwiftWin32/View.swift:270: Fatal error: Unexpectedly found nil while unwrapping an Optional value
Current stack trace:
0    (null)                             0x00007ff8c057c7b0 swift_stdlib_reportFatalErrorInFile + 134
1    (null)                             0x00007ff8c0228ee0 _isStdlibInternalChecksEnabled() + 1822
2    (null)                             0x00007ff8c0228ee0 _isStdlibInternalChecksEnabled() + 1063
3    (null)                             0x00007ff8c0228ee0 _isStdlibInternalChecksEnabled() + 175
4    (null)                             0x00007ff8c0228960 _assertionFailure(_:_:file:line:flags:) + 365
5    (null)                             0x00007ff70d1272e0 View.init(frame:class:style:parent:) + 5175
6    (null)                             0x00007ff70d0f45f0 Control.init(frame:class:style:parent:) + 517
7    (null)                             0x00007ff70d11eca0 TextField.init(frame:) + 485
8    (null)                             0x00007ff70d11ec30 TextField.__allocating_init(frame:) + 92
9    (null)                             0x00007ff70d142b60 ViewTests.testConvertSameParent() + 362
10   (null)                             0x00007ff70d1424a0 __allTests_CoreGraphicsTests() + 795
11   (null)                             0x00007ff70d0222a0 type metadata accessor for AutoLayoutTests + 889
12   (null)                             0x00007ff70d142800 __allTests_ViewTests() + 356
13   (null)                             0x00007ff8e4a81ee0 type metadata accessor for XCTestCase + 427
14   (null)                             0x00007ff8e4a81ee0 type metadata accessor for XCTestCase + 526
15   (null)                             0x00007ff8e4a81ee0 type metadata accessor for XCTestCase + 402
17   (null)                             0x00007ff8e4a80bc0 XCTestCase.invokeTest() + 356
18   (null)                             0x00007ff8e4a807c0 XCTestCase.perform(_:) + 168
19   (null)                             0x00007ff8e4a84ab0 XCTest.run() + 149
20   (null)                             0x00007ff8e4a82b00 XCTestSuite.perform(_:) + 151
21   (null)                             0x00007ff8e4a84ab0 XCTest.run() + 149
22   (null)                             0x00007ff8e4a82b00 XCTestSuite.perform(_:) + 151
23   (null)                             0x00007ff8e4a84ab0 XCTest.run() + 149
24   (null)                             0x00007ff8e4a82b00 XCTestSuite.perform(_:) + 151
25   (null)                             0x00007ff8e4a84ab0 XCTest.run() + 149
26   (null)                             0x00007ff8e4a7f000 XCTMain(_:arguments:observers:) + 1140
27   (null)                             0x00007ff8e4a7ef20 XCTMain(_:) + 109
28   (null)                             0x00007ff70d1429f0 main + 367
29   (null)                             0x00007ff70d1436e0 type metadata accessor for ViewTests + 1220
30   (null)                             0x00007ff90dd47c10 BaseThreadInitThunk + 20
31   (null)                             0x00007ff90fced700 RtlUserThreadStart + 33
```

compnerd:
I took at a stab at a smaller subset of this as I wanted to investigate if a thought that I had to test this stuff would work.  I think that I have something that works with a basic test.  I think that there is more that we should do in terms of testing and of course the other operations need to be implemented.  I've uploaded that as a draft PR at #490, let me know what you think.

I did stick with the approach that you had pointed out - computing the transformation by iteratively walking the view hierarchy.  I'm not sure if `MapWindowPoints` takes into account world transformations, so this seemed safer.  I've also accounted for the transforms and other such things in the implementation, so it is more involved, but I think that the logic overall should be somewhat simpler.

Hopefully that can serve as a guide towards finishing up this work.

rjpilgrim:
Your implementation should be more correct, since it is taking into account the bounds and transforms. I will be testing the theoretical transform cases to see if it things line up with UIView. I will remove the fatal error on AffineTransform setting just to perform these tests, and comment out these test cases when I push. Avoiding using a `Window` seems to be allowing the tests to work. I can get to working on this tomorrow.

I now see an easy way to handle the null case for the single implementation (assuming there will be a negation prefix operator on Point): 
```
 return view?.convert(rect, to: self) ?? -self.convert(rect, to: self)
```

I will probably use your `convert(to:) -> Point` to find the new origin in the `Rect` functions as well, though I now realize there is a need to accumulate a scaling AffineTransform to apply to the size of the rectangle as well, just to make things even more complicated.

compnerd:
Awesome!  I think that I'll merge the two base changes (adding `bounds` and `transform`).  The addition of the new operators is perfectly fine, we should be able to add that internally.  Seems like the alternate implementation helped explain what I was trying to convey.  Hmm, did I miss that transform? :-(

rjpilgrim:
I believe your Point implementation is correct, I just realize there is additional work missing from my original implementation to potentially adjust the size for the convert `Rect` functions. I need to do some tests in iOS space to verify this though.

I think it potentially makes sense to use Vector as the return type of `WindowBasedOrigin()`, renaming it `WindowBasedTranslate()`. I believe traditionally the `Point + Point` operation returns a vector (while `Point + Vector` returns `Point`), but I don't see any practical consequences for just using `Point`s universally.

compnerd:
I really like the suggested name . `WindowBasedTranslate` and that does deserve to be a `Vector`.  I'll adjust that.

rjpilgrim:
Doing some light research, I think the transform case is a bit more complicated than I expected. In particular the `frame` property automatically adjusts to being a bounding box of the UIView when transform is adjusted. 

As such, using the frame property when transform is involved gives incorrect results. Instead, I believe the following implementation will work:
```
extension Rect {
  internal func center() -> Point {
    return CGPoint(x: self.origin.x + self.size.width/2,
      y: self.origin.y + self.size.height/2)
  }
}

public func convert(_ point: Point, to view: View?) -> Point {
  ...
  if let superview = self.superview, superview == view {
    return ((point - self.bounds.center()).applying(self.transform) + self.center)
  } else if let superview = view.superview, superview == self {
    return (point - view.center).applying(view.transform.inverted()) + view.bounds.center()
  }
  ...
}
```

compnerd:
> extension Rect {
>   internal func center() -> Point {
>     return CGPoint(x: self.origin.x + self.size.width/2,
>       y: self.origin.y + self.size.height/2)
>   }
> }

This seems a bit more concise:
```swift
extension Rect {
  internal var center: Point {
    Point(x: self.midX, y: self.midY)
  }
}
```

> using the frame property when transform is involved gives incorrect results.
Hmm, I'm not sure I understand the case that goes wrong.  I think that some test cases would be useful (as they can help explain the scenarios).  The proposal for using `center` feels like it should be correct, except for the fact that it does apply an additional layer of translation since the `bounds` are not guaranteed to be `((0,0),(width,height))` but can rather be an arbitrary subset of the view.

rjpilgrim:
Update:

Though I proposed using `Vector` in the convert functions, I now think it is preferable in fact to just use `Point + Point -> Point` throughout the code and avoiding `Vector` entirely. This is mostly motivated by the fact that `CGVector` from UIKit is missing an `applying()`, meaning if convention is fully followed then there is a lot of unnecessary conversion back and from between `Vector` and `Point`.

For example, the implementation I listed before becomes
```
public func convert(_ point: Point, to view: View?) -> Point {
  ...
  if let superview = self.superview, superview == view {
    return Point.zero + ((Point.zero  + (point - self.bounds.center())).applying(self.transform) + self.center)
  } else if let superview = view.superview, superview == self {
    return Point.zero + ((Point.zero  + (point - view.center)).applying(view.transform.inverted()) + view.bounds.center())
  }
  ...
}
```
If there is a need for a way to produce Vectors from Points down the road, then I would suggest that a new operator like `Point +> Point -> Vector` is added.

I have done a few tests on my center based implementation in UIKit, including some where the bounds rectangle is manipulated, and it seems to be passing every case. Testing in Windows requires a frame implementation that adjusts itself with respect to the transform property, which is a bit involved. I will try to show a working test case with an experimental frame implementation (with no Drawing logic), then remove it after tests have shown to pass for the transform cases.

rjpilgrim:
Update: I have reimplemented the convert functions to try to take into account the transforms and bound adjustments in the view hierarchy. There are still two outstanding issues with this implementation, but I am pushing for reference.

Problem 1. The applying() function on Rect is not implemented correctly. From this page: [https://developer.apple.com/documentation/coregraphics/cgrect/1455875-applying](url):

> Because affine transforms do not preserve rectangles in general, this function returns the smallest rectangle that contains the transformed corner points of the rect parameter. If the affine transform t consists solely of scaling and translation operations, then the returned rectangle coincides with the rectangle constructed from the four transformed corners.

I have commented out a test case in my tests that should work if this is reimplemented.

Problem 2: My current approach for taking into account accumulated transforms in the view hierarchy seems to only work for rotation affine transformations. I think I will need to avoid doing the translate calculation and transformation calculation separately. Instead, I will want to create an affine transform for the caller and the target which directly transforms each view to world space.

My analysis of the center property above was incorrect. This property is special because it is not affected by either the transform or bounds adjustments. In this way, it provides a reference point for its parent view as to where to find the view. The information in bounds and transform cam then be taken into account to find the final position of the view.

compnerd:
Would you mind splitting out the change for `Rect` into a separate PR?  I think that we should be able to get that merged and out of the way.

rjpilgrim:
I now have this implementation working for all transform cases. As I though, accumulating a single affine transform to do the `transform` manipulation as well as the translation relative to `bounds` and `center`, all in the correct order, was the way to go.

compnerd:
If there is anything else that can be split out a standalone change, please do so.   Otherwise, a rebase + squash would be nice.  I would appreciate it if you would grab the original change that I did and keep that as a separate change to preserve the authorship information.

Sources/SwiftWin32/Support/AffineTransform+UIExtensions.swift
```diff
@@ -0,0 +1,14 @@
+// Copyright © 2021 Ryan Allan <rjallan@g.ucla.edu>
+// SPDX-License-Identifier: BSD-3-Clause
+
+extension AffineTransform {
+  internal static func * (_ lhs: AffineTransform, _ rhs: AffineTransform)
```
compnerd, line 5: This looks like it is `AffineTransform.concatenating`.  Why not just use the existing function?

rjpilgrim: Good point. I will rework it.

rjpilgrim:
I will cherry pick your branch into mine to preserve your draft.

rjpilgrim:
I have squashed the commits down on top of your draft, with minor reworking to get rid of unnecessary operators. I will still need to uncomment the transform and bounds observers once the implementation is okayed.

Tests/UICoreTests/ViewTests.swift
```diff
@@ -0,0 +1,143 @@
...
+    var rect = grandChild.convert(inputRect, to: childB)
+    XCTAssertEqual(rect.origin.x, 123.17714789769627, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, 30.274792065592464, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width, 18.031110765667435, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, 17.514574532740156, accuracy: accuracy)
+
+    rect = grandChild.convert(inputRect, to: childA)
+    XCTAssertEqual(rect.origin.x, -370.7599647161262, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, -3194.0537428193356, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width,  126.21777535967215, accuracy: accuracy)
```
compnerd, line 128:
```suggestion
    XCTAssertEqual(rect.size.width, 126.21777535967215, accuracy: accuracy)
```

Tests/UICoreTests/ViewTests.swift
```diff
@@ -0,0 +1,143 @@
...
+    XCTAssertEqual(rect.size.height, 315.26234158932266, accuracy: accuracy)
+
+    rect = grandChild.convert(inputRect, from: childA)
+    XCTAssertEqual(rect.origin.x, -130.5701354815033, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, 97.83304592215717, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width,  1.4556799635159337, accuracy: accuracy)
```
compnerd, line 134:
```suggestion
    XCTAssertEqual(rect.size.width, 1.4556799635159337, accuracy: accuracy)
```

compnerd (commented):
This is starting to really come together!  A lot of whitespace nits, which the new lint task should catch in the future.  I think that we should figure out how to handle the the setters for `transform` and `bounds`.  Worst case, we should be able to do `@testable` imports, and do something as an SPI to allow tests to set the property.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -352,9 +370,9 @@ public class View: Responder {
 
   /// The bounds rectangle, which describes the view’s location and size in its
   /// own coordinate system.
-  public var bounds: Rect {
+  public var bounds: Rect /*{
     didSet { fatalError("\(#function) not yet implemented") }
-  }
+  }*/
```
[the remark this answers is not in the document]
rjpilgrim: Yes, this property is only looked at by my convert functions currently.

Sources/SwiftWin32/Views and Controls/View.swift
```diff
@@ -366,9 +384,9 @@ public class View: Responder {
 
   /// Specifies the transform applied to the view, relative to the center of its
   /// bounds.
-  public var transform: AffineTransform = .identity {
+  public var transform: AffineTransform = .identity /*{
     didSet { fatalError("\(#function) not yet implemented") }
-  }
+  }*/
```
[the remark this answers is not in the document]
rjpilgrim: Same as with bounds.

Sources/SwiftWin32/Views and Controls/View.swift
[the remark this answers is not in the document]
compnerd: If they are read-only properties, then leaving the `didSet` observers in place seems like it should be okay?  The problem is that they are used in testing.  I wonder if we can disable the `fatalError` only in `@testable` builds.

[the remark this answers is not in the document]
rjpilgrim: I am not sure if detection of @testable is possible within the file. The best I could come up with is:
```
didSet { 
  #if DEBUG
  #else
    fatalError("\(#function) not yet implemented") 
  #endif
}
```
with the following added to `swiftSettings` on the SwiftWin32 target in Package.swift:
```
.define("DEBUG", .when(configuration: .debug))
```
This allows swift test to access the property, but it isn't perfect, because there are other contexts where the debug configuration could be used.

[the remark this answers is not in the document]
compnerd: `DEBUG` is actually already defined by SPM.  However, I think that we should instead require that the user explicitly opt into this when running tests.  That is, we should invert this to something like:

```swift
#if !ENABLE_TESTING
fatalError("\(#function) not yet implemented")
#endif
```

When running tests, we could then just do `swift test -DENABLE_TESTING`.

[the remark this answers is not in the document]
rjpilgrim: Yes this works. Though I think the syntax is `swift test -Xswiftc -DENABLE_TESTING`.

rjpilgrim:
Added check on !ENABLE_TESTING, this is expected to fail tests until define is added to swift test call. Commit hash `e6bcec0` accidently had the didSet redeclared.

compnerd:
This looks almost perfect!  I'm going to merge my draft PR, and then merge main into this branch and squash and merge which is about CI history.  But the change itself looks good.  Thank you for sticking through with this.  The final version seems far easier to reason about.

PR #501, opened 2021-05-06
Fix midpoint calculations on Rect

rjpilgrim:
Simple change to fix midpoints on Rect.


compnerd (approved):
I, uh, thats what I get for being dumb and not trying to get tests setup earlier.  Thank you.

PR #503, opened 2021-05-06
Fix applying() function on Rect

rjpilgrim:
This should fix the affine transform operation on Rect.

### DIFF
--- a/Sources/SwiftWin32/CG/Rect.swift
+++ b/Sources/SwiftWin32/CG/Rect.swift
@@ -22,9 +22,20 @@ public struct Rect {
               size: Size(width: Double(width), height: Double(height)))
   }
 
-  func applying(_ transform: AffineTransform) -> Rect {
-    return Rect(origin: self.origin.applying(transform),
-                size: self.size.applying(transform))
+  public func applying(_ transform: AffineTransform) -> Rect {
+    let points: [Point] = [
+      self.origin,
+      self.origin + Point(x: self.size.width, y: 0),
+      self.origin + Point(x: 0, y: self.size.height),
+      self.origin + Point(x: self.size.width, y: self.size.height),
+    ].map { $0.applying(transform) }
+
+    let xs: [Double] = points.map { $0.x }
+    let ys: [Double] = points.map { $0.y }
+    
+    return Rect(origin: Point(x: xs.min()!, y: ys.min()!),
+                size: Size( width: xs.max()! - xs.min()!,
+                            height: ys.max()! - ys.min()!))
   }
 }
 
--- a/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
+++ b/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
@@ -37,7 +37,6 @@ final class CoreGraphicsTests: XCTestCase {
                         size: Size(width: 16, height: 16))
     XCTAssertEqual(r2.midX, 12)
     XCTAssertEqual(r2.midY, 12)
-
   }
 
   func testRectApplyAffineTransform() {
--- a/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
+++ b/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
@@ -42,6 +42,17 @@ final class CoreGraphicsTests: XCTestCase {
 
   func testRectApplyAffineTransform() {
     var rect: Rect =
+        Rect(origin: Point(x: -6, y: -7),
+             size: Size(width: 12, height: 14))
+            .applying(AffineTransform(rotationAngle: Double.pi/2))
+
+    XCTAssertEqual(rect.origin.x, -7)
+    XCTAssertEqual(rect.origin.y, -6)
+    XCTAssertEqual(rect.size.width, 14)
+    XCTAssertEqual(rect.size.height, 12)
+
+
+    rect =
         Rect(origin: Point(x: 45, y: 115),
              size: Size(width: 13, height: 14))
             .applying(AffineTransform(rotationAngle: 42))
--- a/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
+++ b/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
@@ -80,7 +80,6 @@ final class CoreGraphicsTests: XCTestCase {
     XCTAssertEqual(rect.origin.y, -428.36482622296273, accuracy: accuracy)
     XCTAssertEqual(rect.size.width, 117.68398448828839, accuracy: accuracy)
     XCTAssertEqual(rect.size.height, 87.18621695814943, accuracy: accuracy)
-
   }
 
   static var allTests = [
--- a/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
+++ b/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
@@ -37,10 +37,45 @@ final class CoreGraphicsTests: XCTestCase {
                         size: Size(width: 16, height: 16))
     XCTAssertEqual(r2.midX, 12)
     XCTAssertEqual(r2.midY, 12)
+
+  }
+
+  func testRectApplyAffineTransform() {
+    var rect: Rect =
+        Rect(origin: Point(x: 45, y: 115),
+             size: Size(width: 13, height: 14))
+            .applying(AffineTransform(rotationAngle: 42))
+
+    // Test data comes from the UIKit implementation, running on an iPhone,
+    // and so deviation is expected.
+    var accuracy = 1e-13
+                     
+    XCTAssertEqual(rect.origin.x, 82.20082974097352, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, -104.75635541260408, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width, 18.031110765667435, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, 17.514574532740156, accuracy: accuracy)
+
+    // Deviation increases with more operations
+    accuracy = 1e-12
+
+    rect =
+        Rect(origin: Point(x: 45, y: 115),
+             size: Size(width: 13, height: 14))
+            .applying(AffineTransform(rotationAngle: -104))
+            .applying(AffineTransform(translationX: -26, y: 22))
+            .applying(AffineTransform(scaleX: 7, y: 5))
+
+    XCTAssertEqual(rect.origin.x, -856.8534424207578, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, -428.36482622296273, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width, 117.68398448828839, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, 87.18621695814943, accuracy: accuracy)
+
   }
 
   static var allTests = [
     ("testAffineTransformIdentity", testAffineTransformIdentity),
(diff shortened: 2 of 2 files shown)


Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -28,6 +28,25 @@ final class CoreGraphicsTests: XCTestCase {
     XCTAssertFalse(SwiftWin32.AffineTransform().isIdentity)
   }
 
+  func testRectApplyAffineTransform() {
+    let inputRect = Rect(origin: Point(x:45,y:115)
+        , size: Size(width:13, height:14))
+    let transform = AffineTransform(rotationAngle: 42)
+    let rect = inputRect.applying(transform)
+
+    let accuracy = 0.00000000001
```
compnerd, line 37: Why this specific value for the accuracy?

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -28,6 +28,25 @@ final class CoreGraphicsTests: XCTestCase {
...
+
+    let expectedX = 82.20082974097352
+    let expectedY = -104.75635541260408
+    let expectedWidth = 18.031110765667435
+    let expectedHeight = 17.514574532740156
+
+    XCTAssertEqual(rect.origin.x, expectedX, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, expectedY, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width, expectedWidth, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, expectedHeight, accuracy: accuracy)
```
compnerd, line 47:
```suggestion
    XCTAssertEqual(rect.origin.x, 82.20082974097352, accuracy: accuracy)
    XCTAssertEqual(rect.origin.y, -104.75635541260408, accuracy: accuracy)
    XCTAssertEqual(rect.size.width, 18.031110765667435, accuracy: accuracy)
    XCTAssertEqual(rect.size.height, 17.514574532740156, accuracy: accuracy)
```

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -28,6 +28,25 @@ final class CoreGraphicsTests: XCTestCase {
     XCTAssertFalse(SwiftWin32.AffineTransform().isIdentity)
   }
 
+  func testRectApplyAffineTransform() {
+    let inputRect = Rect(origin: Point(x:45,y:115)
+        , size: Size(width:13, height:14))
+    let transform = AffineTransform(rotationAngle: 42)
+    let rect = inputRect.applying(transform)
```
compnerd, line 35:
```suggestion
    let rect: Rect =
        Rect(origin: Point(x: 45, y: 115),
             size: Size(width: 13, height: 14))
            .applying(AffineTransform(rotationAngle: 42))
```

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -22,9 +22,43 @@ public struct Rect {
...
-    return Rect(origin: self.origin.applying(transform),
-                size: self.size.applying(transform))
+  public func applying(_ transform: AffineTransform) -> Rect {
+    var array: [Point] = []
+    array.append((self.origin + Point(x: self.size.width, y: 0))
+        .applying(transform))
+    array.append((self.origin + Point(x: 0, y: self.size.height))
+        .applying(transform))
+    array.append((self.origin + Point(x: self.size.width, y: self.size.height))
+        .applying(transform))
```
compnerd, line 32:
```suggestion
    let points: [Point] = [
      (self.origin + Point(x: self.size.width, y: 0).applying(transform),
      (self.origin + Point(x: 0, y: self.size.height).applying(transform),
      (self.origin + Point(x: self.size.width, y: self.size.height).applying(transform),
    ]
```

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -22,9 +22,43 @@ public struct Rect {
...
+        maxY = i.y
+      }
+    }
+
+    let newHeight = maxY - minY
+    let newWidth = maxX - minX
+    let newOrigin = Point(x:minX, y: minY)
+
+    return Rect(origin: newOrigin,
+                size: Size(width: newWidth, height: newHeight))
```
compnerd, line 61:
```suggestion
    return Rect(origin: Point(x: minX, y: minY,
                size: Size(width: maxX - minX, height: maxY - minY))
```

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -22,9 +22,43 @@ public struct Rect {
...
+      if i.x > maxX {
+        maxX = i.x
+      }
+      if i.y < minY {
+        minY = i.y
+      }
+      if i.y > maxY {
+        maxY = i.y
+      }
+    }
```
compnerd, line 54:
```suggestion
    points.forEach { point in
      minX = min(minX, point.x)
      maxX = max(maxX, point.x)
      minY = min(minY, point.y)
      maxY = max(maxY, point.y)
    }
```

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
rjpilgrim: That was the deviation I was seeing in practice between the UIView.convert() method running on MacOS and my convert() method in my test cases. Testing again, it looks like the deviation is less for this function, which makes sense because there are fewer arithmetic operations. I will instead use `0.0000000000001`. I will also use the scientific notation literal instead: `1e-13`.

rjpilgrim:
I have updated with the more concise implementation and a change of accuracy parameter in the test.

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
compnerd: I suppose I should have been more clear, how can we derive the value for the deviation instead?  Can we just use `.ulp` instead?  Or perhaps we should change the constants such that we can avoid the deviation entirely.  The problem is that the hardware can change the results (e.g. the value may be different on ARM64 or RISC-V).

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -22,9 +22,29 @@ public struct Rect {
...
+
+    points.forEach { point in
+      minX = min(minX, point.x)
+      maxX = max(maxX, point.x)
+      minY = min(minY, point.y)
+      maxY = max(maxY, point.y)
+    }
+
+    return Rect(origin: Point(x: minX, y: minY),
+                size: Size(width: maxX - minX, height: maxY - minY))
```
compnerd, line 47: Sorry, now that there is much less noise, this seems far easier to follow what it is trying to do.  I think that this can be even more concise:

```suggestion
    let points: [Point] = [
      self.origin,
      self.origin + Point(x: self.size.width, y: 0),
      self.origin + Point(x: 0, y: self.size.height),
      self.origin + Point(x: self.size.width, y: self.size.height),
    ].map { $0.applying(transform) }

    let xs: [Double] = points.map { $0.x }
    let ys: [Double] = points.map { $0.y }
    
    return Rect(origin: Point(x: xs.min(), y: ys.min()),
                size: Size(width: xs.max() - xs.min(), height: ys.max() - ys.min())
```

I want to think a little bit more about this, because this seems like it could be even simpler.  But, this definitely is far easier to follow IMO: you have the four points defining the rectangle, which you transform given a transformation.  You then create the rectangle from the minimum x, minimum y, and and the difference of the maximum and minimum xs and ys.  Given that there are 4 points, known statically, this seems like this should be something that should be possible to implement without branching.

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -37,10 +37,27 @@ final class CoreGraphicsTests: XCTestCase {
...
+    XCTAssertEqual(rect.size.width, 18.031110765667435, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, 17.514574532740156, accuracy: accuracy)
   }
 
   static var allTests = [
     ("testAffineTransformIdentity", testAffineTransformIdentity),
-    ("testAffineTransformIdentityIsIdentity", testAffineTransformIdentityIsIdentity)
+    ("testAffineTransformIdentityIsIdentity", testAffineTransformIdentityIsIdentity),
+    ("testRectComputedProperties", testRectComputedProperties),
+    ("testRectApplyAffineTransform", testRectApplyAffineTransform)
```
compnerd, line 61:
```suggestion
    ("testRectApplyAffineTransform", testRectApplyAffineTransform),
```

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
rjpilgrim: In the case of my tests, the deviation is a result of the hardware differences. The test cases come from the UIKit implementation running on an Arm64 iPhone simulator. Running this implementation of applying() as an extension on UIView on the iPhone simulator, there is no deviation at all.

As long as I use the UIView implementation as the baseline, there will be expected deviation happening in the hardware. I do not think these deviations of this scale will have consequences at the rendering stage.

compnerd: Oh, this is not a concern about rendering, its about having an understanding for what is going on, and how to deal with the differences when they come up.  Using values which prevent the rounding issues means that its easier to spot.  I suppose that adding a note that the deviation is for FP64 vs FP80 is a good enough away to deal with that for now, but making it more a value that is derived would be nice.

compnerd:
This is starting to look pretty great.  The one question Im left wondering: for the rotations, would it make sense to use a value that's easier to reason about (e.g. `Double.pi / 4` radians, and then show what the expected math would look like.  At that point, just performing the straight line math and comparing the results would validate the results, and avoid the need for referencing the platform or providing an accuracy.  If that seems like it doesn't actually validate the results well enough, lets merge this to get unblocked.

rjpilgrim:
I personally prefer to use the baseline for testing, but I do see the benefit of having clear and explainable tests.

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -37,10 +37,56 @@ final class CoreGraphicsTests: XCTestCase {
                         size: Size(width: 16, height: 16))
     XCTAssertEqual(r2.midX, 12)
     XCTAssertEqual(r2.midY, 12)
+
```
compnerd, line 40:
```suggestion
```

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -37,10 +37,56 @@ final class CoreGraphicsTests: XCTestCase {
...
+             size: Size(width: 13, height: 14))
+            .applying(AffineTransform(rotationAngle: -104))
+            .applying(AffineTransform(translationX: -26, y: 22))
+            .applying(AffineTransform(scaleX: 7, y: 5))
+
+    XCTAssertEqual(rect.origin.x, -856.8534424207578, accuracy: accuracy)
+    XCTAssertEqual(rect.origin.y, -428.36482622296273, accuracy: accuracy)
+    XCTAssertEqual(rect.size.width, 117.68398448828839, accuracy: accuracy)
+    XCTAssertEqual(rect.size.height, 87.18621695814943, accuracy: accuracy)
+
```
compnerd, line 83:
```suggestion
```

compnerd (approved):

PR #506, opened 2021-05-06
Add Rect center computed property

rjpilgrim:
Adding helper functions to `Rect`

### DIFF
--- a/Sources/SwiftWin32/Support/Rect+UIExtensions.swift
+++ b/Sources/SwiftWin32/Support/Rect+UIExtensions.swift
@@ -19,3 +19,9 @@ extension RECT {
               bottom: LONG(from.origin.y + from.size.height))
   }
 }
+
+extension Rect {
+  internal var center: Point {
+    Point(x: self.midX, y: self.midY)
+  }
+}


Sources/SwiftWin32/Support/Rect+UIExtensions.swift
```diff
@@ -19,3 +19,19 @@ extension RECT {
...
+}
+
+extension Rect {
+  internal static func + (_ lhs: Rect, _ rhs: Point) -> Rect {
+    return Rect(origin: Point(x: lhs.origin.x + rhs.x, y: lhs.origin.y + rhs.y), size: lhs.size)
+  }
+  
+  internal static func - (_ lhs: Rect, _ rhs: Point) -> Rect {
+    return Rect(origin: Point(x: lhs.origin.x - rhs.x, y: lhs.origin.y - rhs.y), size: lhs.size)
+  }
```
compnerd, line 36: I don't think that these functions make sense.  What does it mean mathematically to subtract a point from a rectangle?  I think that #508 adds the `offsetBy(dx:dy:)` method to `Rect` which is an expected public method.  That should allow you to implement these inline with a splat of the point.

compnerd (changes requested):

Sources/SwiftWin32/Support/Rect+UIExtensions.swift
rjpilgrim: Good point. I will remove that from this pull request.

compnerd (approved):

PR #512, opened 2021-05-08
Rect: refine `Rect.applying(_:)`

compnerd:
Use a bit more functional style computation in the application of a
transformation.  Use a single pass to simultaneously compute the new
bounds of the rectangle.


compnerd:
@rjpilgrim I think your intuition of the tuple might have been right, and that this might be a nicer way to describe the transformation.

rjpilgrim:
Yes, I like this implementation. Very Haskell-like.

compnerd:
Haha, yes, I find that I do end up with some haskell-ish Swift.

PR #516, opened 2021-05-09
Rect: add additional `Rect` interfaces

compnerd:
This adds the `standardized`, `integeral`, `isEmpty` properties, the
`insetBy` method and fixes an oversight in the `applying` method.


compnerd:
CC: @rjpilgrim

I was implementing some of the other interfaces, and realized that the implementation of `applying(_:)`  was incorrect for non-standard form rectangles.  There is a new test case for that now.  Turns out, the micro-optimization that I was trying to get away with needs a bit more help as well (the identity transformation still needs to standardize the rectangle).  There was a small note in the reference platform documentation indicating that null rectangles are handled specially.

PR #521, opened 2021-05-11
Create documentation.yml

compnerd:
Add a new CI task for auto-generating documentation content.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,61 @@
+name: documentation
+
+on:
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @yostane - btw, this is roughly what the auto-generation step is that I was referring to.  It will auto-trigger on a merge to main, regenerating the documentation, re-creating the gh-pages branch.  We should be able to merge in the content from the website branch into root.  If we need to, we should be able to shuffle the documentation down a directory.

PR #531, opened 2021-05-13
CoreGraphics: add support for intersection

compnerd:
This adds the `intersection` and `intersects` methods to `Rect`.


compnerd:
CC: @rjpilgrim

PR #532, opened 2021-05-14
Add contains() functions to Rect

rjpilgrim:
Adding the functions `contains()` functions to `Rect`, testing if a `Rect` contains a given `Point` or different `Rect`. The `Rect` case is implemented with two calls to the `Point` function for both the top left and bottom right corners.

### DIFF
--- a/Sources/SwiftWin32/CG/Rect.swift
+++ b/Sources/SwiftWin32/CG/Rect.swift
@@ -211,6 +211,19 @@ public struct Rect {
     return !intersection(rect).isEmpty
   }
 
+  /// Returns whether a rectangle contains a specified point.
+  public func contains(_ point: Point) -> Bool {
+    guard !self.isNull else { return false }
+    let standardized: Rect = self.standardized
+    return standardized.minX...standardized.maxX ~= point.x
+        && standardized.minY...standardized.maxY ~= point.y
+  }
+
+  /// Returns whether the first rectangle contains the second rectangle.
+  public func contains(_ rect2: Rect) -> Bool {
+    return self == self.union(rect2)
+  }
+
   /// Returns whether a rectangle has zero width or height, or is a null
   /// rectangle.
   public var isEmpty: Bool {
--- a/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
+++ b/Tests/CoreGraphicsTests/CoreGraphicsTests.swift
@@ -230,6 +230,43 @@ final class CoreGraphicsTests: XCTestCase {
     XCTAssertEqual(Rect.null.union(Rect.infinite), Rect.infinite)
   }
 
+  func testRectContains() {
+    let r1: Rect = Rect(x: 0, y: 0, width: 100, height: 100)
+    let r2: Rect = Rect(x: 25, y: 25, width: 50, height: 50)
+    let r3: Rect = Rect(x: 75, y: 75, width: 50, height: 50)
+    let r4: Rect = Rect(x: 125, y: 125, width: 50, height: 50)
+    let r5: Rect = Rect(x: 75, y: 75, width: -50, height: -50)
+
+
+    XCTAssertTrue(r1.contains(r2))
+    XCTAssertFalse(r2.contains(r1))
+
+    XCTAssertFalse(r1.contains(r3))
+    XCTAssertFalse(r3.contains(r1))
+
+    XCTAssertFalse(r1.contains(r4))
+    XCTAssertFalse(r4.contains(r1))
+
+    XCTAssertTrue(r1.contains(r5))
+    XCTAssertFalse(r5.contains(r1))
+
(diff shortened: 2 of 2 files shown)


Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -193,6 +193,24 @@ public struct Rect {
     return !intersection(rect).isEmpty
   }
 
+  /// Returns whether a rectangle contains a specified point.
+  public func contains(_ point: Point) -> Bool {
+    guard !self.isNull else { return false }
+    let standardized: Rect = self.standardized
+    return standardized.minX...standardized.maxX ~= point.x
+        && standardized.minY...standardized.maxY ~= point.y
+        ? true : false
```
compnerd, line 202:
```suggestion
```
The value itself is already a boolean, and in the correct state, so we don't need the ternary

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -193,6 +193,24 @@ public struct Rect {
     return !intersection(rect).isEmpty
   }
 
+  /// Returns whether a rectangle contains a specified point.
+  public func contains(_ point: Point) -> Bool {
+    guard !self.isNull else { return false }
+    let standardized: Rect = self.standardized
+    return standardized.minX...standardized.maxX ~= point.x
+        && standardized.minY...standardized.maxY ~= point.y
```
compnerd, line 201: I wonder if this is better than the explicit form.  I don't think that the micro-optimization is important right now, something to consider for the future.

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -193,6 +193,24 @@ public struct Rect {
...
+        ? true : false
+  }
+
+  /// Returns whether the first rectangle contains the second rectangle
+  public func contains(_ rect2: Rect) -> Bool {
+    guard !self.isNull, !rect2.isNull else { return false }
+    let lhs: Rect = self.standardized, rhs: Rect = rect2.standardized
+    return lhs.contains(rhs.origin)
+        && lhs.contains(Point(x: rhs.maxX, y: rhs.maxY))
+        ? true : false
```
compnerd, line 211: Why not just do `self.union(rect2).isEmpty`?

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -203,6 +203,26 @@ final class CoreGraphicsTests: XCTestCase {
...
+
+    XCTAssertFalse(r1.contains(r3))
+    XCTAssertFalse(r3.contains(r1))
+
+    XCTAssertFalse(r1.contains(r4))
+    XCTAssertFalse(r4.contains(r1))
+
+    XCTAssertTrue(r1.contains(r5))
+    XCTAssertFalse(r5.contains(r1))
+  }
```
compnerd, line 224: Does it make sense to have some test cases for `null`, `empty`, and `zero` rects?

Sources/SwiftWin32/CG/Rect.swift
rjpilgrim: I went ahead and reimplemented by calling union.

Sources/SwiftWin32/CG/Rect.swift
rjpilgrim: A silly mistake on my part.

Tests/CoreGraphicsTests/CoreGraphicsTests.swift
rjpilgrim: I added a test case for null. Perhaps I should add another.

rjpilgrim:
I reworked the implementation of `contains(Rect)` using union as you suggested. As part of this rework, I noticed that the equality operator needed a special implementation, because the UIKit implementation does in fact return true if the two rects are equal after standardization. This removes the need to do a standardization at the top of that function.

compnerd:
Interesting find on the equality!  Mind splitting that out into a separate change?

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -173,6 +173,20 @@ public struct Rect {
                 height: standardized.size.height)
   }
 
+  /// Returns the smallest rectangle that contains the two source rectangles
+  public func union(_ r2: Rect) -> Rect {
```
compnerd, line 177: Oh sorry, I didn't realize that I hadn't pushed my changes for `union`.  I have #533 which has the changes that I had sitting around locally.  I think that the implementation is pretty similar.  The test cases are different though, and I think that mine might be a bit more exhaustive.

rjpilgrim:
Rebased `contains()` change on top of current main, with non standard equality fix.

compnerd (approved):

PR #534, opened 2021-05-14
Rework == operator on Rect to take standardization into account

rjpilgrim:
A simple test showing the operator at work is also included..

### DIFF
--- a/Sources/SwiftWin32/CG/Rect.swift
+++ b/Sources/SwiftWin32/CG/Rect.swift
@@ -206,11 +206,23 @@ public struct Rect {
 
   /// Returns whether the rectangle is equal to the null rectangle.
   public var isNull: Bool {
-    return self == .null
+    return self.pureEqualityCheck(.null)
+  }
+
+  @inline(__always) private func pureEqualityCheck(_ rhs: Rect) 
+      -> Bool {
+    return (self.origin == rhs.origin) && (self.size == rhs.size)
   }
 }
 
+
+
 extension Rect: Equatable {
+  // MARK - Operator Functions
+  public static func == (lhs: Rect, rhs: Rect) -> Bool {
+    let lhs: Rect = lhs.standardized, rhs: Rect = rhs.standardized
+    return lhs.pureEqualityCheck(rhs)
+  }
 }
 
 extension Rect: CustomDebugStringConvertible {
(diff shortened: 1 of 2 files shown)


Tests/CoreGraphicsTests/CoreGraphicsTests.swift
```diff
@@ -203,6 +203,13 @@ final class CoreGraphicsTests: XCTestCase {
     XCTAssertFalse(r4.intersects(r1))
   }
 
+  func testRectNonstandardEquality() {
+    let rectOne: Rect = Rect(x: 0, y: 0, width: 10, height: 10)
+    let rectTwo: Rect = Rect(x: 10, y: 10, width: -10, height: -10)
+
+    XCTAssertTrue(rectOne == rectTwo)
+  }
```
compnerd, line 211: Better test coverage for this would be nice, but that can be a follow up.

compnerd (approved):

compnerd:
Interesting, this seems to be causing a crash somehow?

Sources/SwiftWin32/CG/Rect.swift
```diff
@@ -211,6 +211,12 @@ public struct Rect {
 }
 
 extension Rect: Equatable {
+  // MARK - Operator Functions
+  public static func == (lhs: Rect, rhs: Rect) -> Bool {
+    guard !lhs.isNull, !rhs.isNull else { return false }
```
compnerd, line 216: Actually, this seems wrong .. the null rect should be equal no?

rjpilgrim: Good point, that line can just go away.

rjpilgrim:
> Interesting, this seems to be causing a crash somehow?

Yes, but I do not understand it. Commenting out the entire new function for `==` causes everything to work again, with the my new test failing as expected.

compnerd:
Its unclear why it is failing, would require running under a debugger.

rjpilgrim:
> ~Its unclear why it is failing, would require running under a debugger.~ Its infinite recursion, there is a stack overflow (`isNull` invokes `==`, invokes `isNull`, repeat ad-nauseum)

And Standardized calls isNull... I think there is need for a private non-standardized equality function just to implement isNull.

compnerd:
> > ~Its unclear why it is failing, would require running under a debugger.~ Its infinite recursion, there is a stack overflow (`isNull` invokes `==`, invokes `isNull`, repeat ad-nauseum)
> 
> And Standardized calls isNull... I think there is need for a private non-standardized equality function just to implement isNull.

Sure, I don't have an idea off the top of my head how to break the cycle, so restructuring with a `@inline(__always) private` function should be fine.

rjpilgrim:
Test is now running correctly.

PR #542, opened 2021-05-17
Add an instruction for running unit tests

tomaszpieczykolan:
Hi :)

I would like to contribute to Swift/Win32 a little bit  - I want to make some classes `open`.

I first wanted to run the unit tests to make sure I don't break anything, but I struggled to find out how to do it. I'm used to programming on macOS and doing it on Windows is a little bit confusing to me. Anyways, after few hours of digging through Google and this repo I finally found the solution and I thought it would be great if it could be documented in the readme.

I am not sure if those instructions are missing something, but in my case this was enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,3 @@ copy Examples\UICatalog\Info.plist .build\x86_64-unknown-windows-msvc\debug\
 In order to run unit tests of Swift/Win32 you need to:
 1. Open `x64 Native Tools Command Prompt for VS 2019`.
 2. Run `swift test -Xswiftc -DENABLE_TESTING`.
-
-### Required software for running tests
-
-| Component | Visual Studio ID |
-| --- | --- |
-| Git for Windows | Microsoft.VisualStudio.Component.Git |
--- a/README.md
+++ b/README.md
@@ -48,8 +48,7 @@ copy Examples\UICatalog\Info.plist .build\x86_64-unknown-windows-msvc\debug\
 
 In order to run unit tests of Swift/Win32 you need to:
 1. Open `x64 Native Tools Command Prompt for VS 2019` as administrator.
-2. `cd` to the location of Swift/Win32 on your local computer.
-3. Run `swift test -Xswiftc -DENABLE_TESTING`.
+2. Run `swift test -Xswiftc -DENABLE_TESTING`.
 
 ### Required software for running tests
 
--- a/README.md
+++ b/README.md
@@ -43,3 +43,16 @@ mt -nologo -manifest Examples\UICatalog\UICatalog.exe.manifest -outputresource:.
 copy Examples\UICatalog\Info.plist .build\x86_64-unknown-windows-msvc\debug\
 .build\x86_64-unknown-windows-msvc\debug\UICatalog.exe
 ```
+
+## Running tests
+
+In order to run unit tests of Swift/Win32 you need to:
+1. Open `x64 Native Tools Command Prompt for VS 2019` as administrator.
+2. `cd` to the location of Swift/Win32 on your local computer.
+3. Run `swift test -Xswiftc -DENABLE_TESTING`.
+
+### Required software for running tests
+
+| Component | Visual Studio ID |
+| --- | --- |
+| Git for Windows | Microsoft.VisualStudio.Component.Git |
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ copy Examples\UICatalog\Info.plist .build\x86_64-unknown-windows-msvc\debug\
 ## Running tests
 
 In order to run unit tests of Swift/Win32 you need to:
-1. Open `x64 Native Tools Command Prompt for VS 2019` as administrator.
+1. Open `x64 Native Tools Command Prompt for VS 2019`.
 2. Run `swift test -Xswiftc -DENABLE_TESTING`.
 
 ### Required software for running tests


README.md
```diff
@@ -43,3 +43,16 @@ mt -nologo -manifest Examples\UICatalog\UICatalog.exe.manifest -outputresource:.
 copy Examples\UICatalog\Info.plist .build\x86_64-unknown-windows-msvc\debug\
 .build\x86_64-unknown-windows-msvc\debug\UICatalog.exe
 ```
+
+## Running tests
+
+In order to run unit tests of Swift/Win32 you need to:
+1. Open `x64 Native Tools Command Prompt for VS 2019` as administrator.
```
compnerd, line 50: This shouldn't be necessary.  The use of the administrator shell is definitely incorrect.  Nothing should require the elevated privileges.

README.md
```diff
@@ -43,3 +43,16 @@ mt -nologo -manifest Examples\UICatalog\UICatalog.exe.manifest -outputresource:.
...
+2. `cd` to the location of Swift/Win32 on your local computer.
```
compnerd, line 51: I think that having something more copy-and-paste friendly assuming that the user is in the root of the tree is better.  See above for the example of building.

README.md
```diff
@@ -43,3 +43,16 @@ mt -nologo -manifest Examples\UICatalog\UICatalog.exe.manifest -outputresource:.
...
+3. Run `swift test -Xswiftc -DENABLE_TESTING`.
```
compnerd, line 52: Ah, this is a good thing to document.  This was introduced recently by @rjpilgrim to enable testing some additional cases for hit testing work.

README.md
```diff
@@ -43,3 +43,16 @@ mt -nologo -manifest Examples\UICatalog\UICatalog.exe.manifest -outputresource:.
...
+
+### Required software for running tests
+
+| Component | Visual Studio ID |
+| --- | --- |
+| Git for Windows | Microsoft.VisualStudio.Component.Git |
```
compnerd, line 58: I don't think that this section should be included.  Especially when you consider that git isn't actually used by the project - you could just as well as download a snapshot of main as a zip file from the GitHub interface.  That said, git is required for SPM, and is already part of the instructions at swift.org

compnerd (changes requested):

compnerd:
> Anyways, after few hours of digging through Google and this repo I finally found the solution and I thought it would be great if it could be documented in the readme.

BTW, it might've been easier to look at https://github.com/compnerd/swift-win32/blob/main/.github/workflows/build.yml#L49-L50 for how to run the tests.

tomaszpieczykolan:
> > Anyways, after few hours of digging through Google and this repo I finally found the solution and I thought it would be great if it could be documented in the readme.
> 
> BTW, it might've been easier to look at https://github.com/compnerd/swift-win32/blob/main/.github/workflows/build.yml#L49-L50 for how to run the tests.

Yes, I did, and it helped me quite a bit :)

README.md
tomaszpieczykolan: I run into some issues when testing without administrator privileges and run into some issues.

I had this error:
`error: The operation could not be completed. (NSCocoaErrorDomain error 256.)`

README.md
tomaszpieczykolan: Yes, I think it is only required by SPM, and in my case I was unable to run the tests without installing git via Visual Studio Installer.

README.md
tomaszpieczykolan: Hmmm, in this case would you just remove the step 2 completely?

README.md
tomaszpieczykolan: Also, I couldn't find information about git being required by SPM on swift.org. Currently git is only listed as a recommended software, but there is no explanation as to why it is recommended - maybe it makes sense to add this info over there?

README.md
compnerd: Yes

README.md
compnerd: You didn't enable developer mode on WIndows.  Alternatively, you could assign your user `SE_ALLOW_CREATE_SYMBOLIC_LINK` permissions via a GPO (using secpol.msc or gpedit.msc). 

https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/ covers the permissions thing.  I don't think that this belongs in the documentation.

README.md
compnerd: The required, recommended, suggested is a standard categorization.  required - this is mandatory, recommended - this is recommended, failure to install will result in broken functionality, suggested - you likely want to install this, but it shouldn't result in broken functionality.

compnerd:
BTW, I think that converting `ViewController` to `open` is something that would enable me to do another item that I'd like to do, so please do send that change along before this gets completed.

README.md
tomaszpieczykolan: Resolved in 4dbfac0a431f9197e67cb57ac4fdb7ba1832d0fa

README.md
tomaszpieczykolan: Somehow it was not clear for me. Maybe it's just because I'm a newbie, but I think there is value in making things a bit easier for the newbies, even if it means duplicating information. Anyways, I can remove it :)

README.md
tomaszpieczykolan: Not sure if you are asking to "document this line a little bit more", or are you just saying that "it's good that it is documented now"? If it's the latter, it seems that this conversation can be marked as resolved? Otherwise, I'll try to do my best to document why those options are being used here.

README.md
compnerd: I think that I will see if we perhaps we can make that more clear in the documentation at swift.org.  I don't think that it should be something which gets replicated into every single project.

README.md
compnerd: Oh, I meant, that you have a valid point.  The "standard" `swift test` won't work for the project for the time being because we need to have some specific support for testing due to missing functionality.  As such, we should document how to run tests.  Until recently, `swift test` was all that was needed.

README.md
compnerd: Again, the adminsitrator rights are incorrect.  I think that the expectation that developers will enable developer mode on the machine is reasonable, and not something specific to this project.

tomaszpieczykolan: Thanks, I didn't even know that Windows has a developer mode. Should've been obvious to me, since macOS also has it, but it wasn't 😅

Do you think it would be beneficial if it was mentioned on https://swift.org/getting-started/? I think many people using Swift on Windows are going to be people just like me, who don't have a lot of experience developing on Windows, and it might save them a bit of time.

compnerd: Yeah, having that be mentioned there seems reasonable.

README.md
tomaszpieczykolan: Ok, I'm going to mark this conversation as resolved then :)

README.md
tomaszpieczykolan: I removed the "as administrator" part - marking this as resolved.

README.md
tomaszpieczykolan: I removed this section - marking this as resolved.

compnerd (approved):

PR #544, opened 2021-05-18
Make ViewController open

tomaszpieczykolan:
This PR makes the `ViewController` class `open`, so that it can be overriden.

### DIFF
--- a/Tests/UICoreTests/ViewControllerTests.swift
+++ b/Tests/UICoreTests/ViewControllerTests.swift
@@ -39,3 +39,32 @@ final class ViewControllerTests: XCTestCase {
     XCTAssertEqual(sut.title, "Title")
   }
 }
+
+final class MockViewController: ViewController {
+  var modalPresentationStyleGetter: () -> ModalPresentationStyle = {
+    XCTFail("Not implemented")
+    return .automatic
+  }
+
+  var modalPresentationStyleSetter: (ModalPresentationStyle) -> Void = { _ in
+    XCTFail("Not implemented")
+  }
+
+  var viewDidLoadBlock: () -> Void = {
+    XCTFail("Not implemented")
+  }
+
+  override var modalPresentationStyle: ModalPresentationStyle {
+    get {
+      return modalPresentationStyleGetter()
+    }
+    set {
+      modalPresentationStyleSetter(newValue)
+    }
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    viewDidLoadBlock()
+  }
+}
--- a/Tests/UICoreTests/ViewControllerTests.swift
+++ b/Tests/UICoreTests/ViewControllerTests.swift
@@ -6,102 +6,102 @@ import XCTest
 
 final class ViewControllerTests: XCTestCase {
   func testLazyViewLoading() {
-    let sut = ViewController()
+    let vc = ViewController()
 
-    XCTAssertNil(sut.viewIfLoaded)
-    XCTAssertFalse(sut.isViewLoaded)
+    XCTAssertNil(vc.viewIfLoaded)
+    XCTAssertFalse(vc.isViewLoaded)
 
-    _ = sut.view
+    _ = vc.view
 
-    XCTAssertNotNil(sut.viewIfLoaded)
-    XCTAssertTrue(sut.isViewLoaded)
+    XCTAssertNotNil(vc.viewIfLoaded)
+    XCTAssertTrue(vc.isViewLoaded)
   }
 
   func testManualViewLoading() {
-    let sut = ViewController()
+    let vc = ViewController()
 
-    XCTAssertNil(sut.viewIfLoaded)
-    XCTAssertFalse(sut.isViewLoaded)
+    XCTAssertNil(vc.viewIfLoaded)
+    XCTAssertFalse(vc.isViewLoaded)
 
-    sut.loadViewIfNeeded()
+    vc.loadViewIfNeeded()
 
-    XCTAssertNotNil(sut.viewIfLoaded)
-    XCTAssertTrue(sut.isViewLoaded)
+    XCTAssertNotNil(vc.viewIfLoaded)
+    XCTAssertTrue(vc.isViewLoaded)
   }
 
   func testTitleGetterAndSetter() {
-    let sut = ViewController()
+    let vc = ViewController()
 
-    // XCTAssertNil(sut.title) // This is currently failing, but the initial value of `title` should be `nil`
+    // XCTAssertNil(vc.title) // This is currently failing, but the initial value of `title` should be `nil`
 
-    sut.title = "Title"
+    vc.title = "Title"
 
-    XCTAssertEqual(sut.title, "Title")
+    XCTAssertEqual(vc.title, "Title")
   }
 
   func testValueOfDisablesAutomaticKeyboardDismissal() {
-    let sut = MockViewController()
+    let vc = MockViewController()
 
-    sut.modalPresentationStyleGetter = {
+    vc.modalPresentationStyleGetter = {
       return .automatic
     }
 
-    XCTAssertFalse(sut.disablesAutomaticKeyboardDismissal)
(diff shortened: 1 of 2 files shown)


tomaszpieczykolan:
This is work in progress. Before submitting for review, I would like to audit the whole `ViewController` class and write some more unit tests for it.

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,28 @@
+import XCTest
```
compnerd, line 1: Please add a comment header to indicate the license.

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
compnerd, line 1: Mind throwing a copyright line as well?

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+import XCTest
+@testable import SwiftWin32
+
+final class ViewControllerTests: XCTestCase {
+  func testLazyViewLoading() {
+    let sut = ViewController()
```
compnerd, line 8: What is the name `sut` derived from?

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+    sut.loadViewIfNeeded()
+
+    XCTAssertNotNil(sut.viewIfLoaded)
+    XCTAssertTrue(sut.isViewLoaded)
+  }
+
+  func testTitleGetterAndSetter() {
+    let sut = ViewController()
+
+    // XCTAssertNil(sut.title) // This is currently failing, but the initial value of `title` should be `nil`
```
compnerd, line 34: Oh interesting, this sounds like you've found a bug!

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    XCTAssertEqual(sut.title, "Title")
+  }
+
+  func testValueOfDisablesAutomaticKeyboardDismissal() {
+    let sut = ViewController()
+
+    XCTAssertFalse(sut.disablesAutomaticKeyboardDismissal)
+
+    // sut.modalPresentationStyle = .formSheet // This is currently failing due to missing implementation
```
compnerd, line 46: This should be pretty easy to add a stub for.

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+  func testNextResponder() {
+    let expectedResult = View(frame: .zero)
+
+    let sut = ViewController()
+
+    XCTAssertNil(sut.next)
+
+    expectedResult.addSubview(sut.view)
+
+    XCTAssert(sut.next === expectedResult)
```
compnerd, line 60: Can this not use `XCTAssertEqual`?

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    let expectation = self.expectation(description: "viewDidLoad should be called")
+
+    sut.viewDidLoadBlock = {
+      expectation.fulfill()
+    }
+
+    sut.loadViewIfNeeded()
+
+    wait(for: [expectation], timeout: 0.1)
```
compnerd, line 88: Oh, this is a really nice approach!

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    let sut = ViewController()
+
+    XCTAssertNil(sut.next)
+
+    expectedResult.addSubview(sut.view)
+
+    XCTAssert(sut.next === expectedResult)
+  }
+
```
compnerd, line 62: Please fix the lint issue.

Tests/UICoreTests/ViewControllerTests.swift
```diff
@@ -0,0 +1,114 @@
...
+    sut.viewDidLoadBlock = {
+      expectation.fulfill()
+    }
+
+    sut.loadViewIfNeeded()
+
+    wait(for: [expectation], timeout: 0.1)
+  }
+
+func testViewDidLoadMethodNotCalledIfViewNotLoaded() {
```
compnerd, line 91: Please fix the lint issue.

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: "Subject under test"

I'm just used to this from my job 😅

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: Here I'm checking if it's the same object. I could use `XCTAssertEqual`, assuming that the object conforms to `Equatable`, but that wouldn't be a check whether it's the same instance – two different instances can be "equal".

compnerd: Oh, didn't spot the `===` vs `==` in the browser.

Tests/UICoreTests/ViewControllerTests.swift
compnerd: Its not currently used in the project, so it would be nice to retain consistency throughout the rest of the project.

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: Done :)

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: Done

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: It seems that in this project you are using the name of the class? So in this case I would need to rename `sut` to `viewController`?

As far as I know `sut` is a relatively common convention, so I would suggest considering switching to using it. It's a short name, which helps avoid very long lines, and once you get used to it, it's easy to see exactly which object is being tested.

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: Thanks :) I assume this discussion can be marked as resolved 😅

Tests/UICoreTests/ViewControllerTests.swift
tomaszpieczykolan: Can this bug be fixed outside the scope of this PR, and can we mark this discussion as resolved?

Tests/UICoreTests/ViewControllerTests.swift
compnerd: Yeah, `viewController` or `vc` are fine.

Tests/UICoreTests/ViewControllerTests.swift
compnerd: Yeah, that seems reasonable.  Please write the test with the correct expectation; I'll add a utility to help mark the test as expected to fail.

compnerd:
@tomaszpieczykolan would you mind rebasing the change?  I think that the new features that I've enabled will be nice to take advantage of.

egorzhdan:
The ability to subclass `ViewController` & override `buildMenu(with: MenuBuilder)` is really helpful for menus, it would be great to merge this PR if there is no concern left.

compnerd:
I that that when the class is being marked `open`, we need to ensure that the members are properly marked as `open` as well, which is not the case yet.  Furthermore, @tomaszpieczykolan has this marked as a draft still.  I definitely am waiting on this change as I have some changes sitting around that require this to be open.

PR #546, opened 2021-05-18
Update README.md

compnerd:
Add more context around the testing of this package.  This explains why we need the difference and what the difference is.  Additionally, mark the fact that the VS2019 terminal should be used for testing and building with SPM.


compnerd:
CC: @tomaszpieczykolan

I've added some more context around the reason for the changes, hopefully this is just as clear.

tomaszpieczykolan:
Awesome, thanks! It's much easier to understand why that flag exists now.

PR #553, opened 2021-05-22
Menus and Shortcuts: open `MenuElement`

compnerd:
`MenuElement` should be an `open` class so that users may subclass the
`MenuElement` type for implementing custom accessors.


compnerd:
CC: @tomaszpieczykolan

PR #556, opened 2021-05-22
Menus and Shortcuts: open Action for subclassing

compnerd:
`Action` should be an `open` class so that users may subclass it.  The
class should have all the expected properties at least stubbed.


compnerd:
CC: @tomaszpieczykolan

PR #620, opened 2021-06-26
Menus and Shortcuts: add overloads for window menu construction

egorzhdan:
This mimics the UIKit's `UIResponder` API. These overloads will be wired to the `ViewController` to actually display the menus.

By default let's just pass the events to the next responder. The subclasses which need a custom menu will need to override these.


compnerd (approved):
It would be nice to get some additional test coverage for this in the future.

egorzhdan:
Agreed, I just think it will be easier to add some tests after `buildMenu` is actually invoked by the `ViewController` logic.

PR #626, opened 2021-07-01
Menus and Shortcuts: make `_MenuBuilder` materialize the menus

egorzhdan:
This change teaches `_MenuBuilder` to build the window menus. It handles submenus & icons similarly to `Win32Menu`.

### DIFF
--- a/Sources/SwiftWin32/Platform/Win32+Menu.swift
+++ b/Sources/SwiftWin32/Platform/Win32+Menu.swift
@@ -13,7 +13,7 @@ internal final class _MenuBuilder: MenuSystem {
   private var menus: OrderedSet<Menu>
 
   internal init?(for view: View) {
-    self.hMenu = MenuHandle(owning: CreateMenu())
+    self.hMenu = MenuHandle(owning: view is Window ? CreateMenu() : CreatePopupMenu())
     if self.hMenu.value == nil {
       log.warning("CreateMenu: \(Error(win32: GetLastError()))")
       return nil
@@ -22,8 +22,42 @@ internal final class _MenuBuilder: MenuSystem {
     self.menus = []
   }
 
+  private func append<T: Collection>(_ menus: T, to hMenu: MenuHandle)
+      where T.Element: MenuElement {
+    for (index, child) in menus.enumerated() {
+      let hImage: BitmapHandle? =
+          BitmapHandle(from: child.image?.bitmap)
+
+      var hSubmenu: MenuHandle?
+      if let submenu = child as? Menu {
+        hSubmenu = MenuHandle(referencing: CreatePopupMenu())
+        if hSubmenu?.value == nil {
+          log.warning("CreateMenu: \(Error(win32: GetLastError()))")
+          continue
+        }
+        append(submenu.children, to: hSubmenu!)
+      }
+
+      var wide = child.title.wide
+      var info = wide.withUnsafeMutableBufferPointer {
+        MENUITEMINFOW(cbSize: UINT(MemoryLayout<MENUITEMINFOW>.size),
+                      fMask: UINT(MIIM_FTYPE | MIIM_STATE | MIIM_ID | MIIM_STRING | MIIM_SUBMENU | MIIM_DATA | MIIM_BITMAP),
+                      fType: UINT(MFT_STRING), fState: UINT(MFS_ENABLED), wID: UInt32.random(in: .min ... .max),
+                      hSubMenu: hSubmenu?.value, hbmpChecked: nil, hbmpUnchecked: nil, dwItemData: 0,
+                      dwTypeData: $0.baseAddress, cch: UINT(child.title.count), hbmpItem: hImage?.value)
+      }
+      InsertMenuItemW(hMenu.value, UINT(index), true, &info)
+    }
+  }
+
   override func setNeedsRebuild() {
-    // TODO(compnerd) create the actual menus
+    // Remove the old submenus from the menu:
+    for index in (0..<GetMenuItemCount(self.hMenu.value)).reversed() {
+      DeleteMenu(self.hMenu.value, UINT(index), UINT(MF_BYPOSITION))
+    }
+
+    // Create the new submenus and add them to the root menu:
+    append(self.menus, to: self.hMenu)
   }
 }
 
--- a/Tests/UICoreTests/MenuBuilderTests.swift
+++ b/Tests/UICoreTests/MenuBuilderTests.swift
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 import XCTest
+import WinSDK
 @testable import SwiftWin32
 
 final class MenuBuilderTests: XCTestCase {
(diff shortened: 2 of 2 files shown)


Tests/UICoreTests/MenuBuilderTests.swift
```diff
@@ -85,8 +85,14 @@ final class MenuBuilderTests: XCTestCase {
...
-    let builder = _MenuBuilder(for: view)
-    XCTAssertNotNil((builder?.system as? _MenuBuilder)?.hMenu.value)
+    guard let builder = _MenuBuilder(for: view) else { return }
+
+    XCTAssertNotNil((builder.system as? _MenuBuilder)?.build()?.value)
+
+    let menu = Menu(title: "title", children: [Command(title: "cmd") { _ in }])
+    builder.insertSibling(menu, afterMenu: .root)
+
+    XCTAssertNotNil((builder.system as? _MenuBuilder)?.build()?.value)
```
egorzhdan, line 95: I don't really know a good way to test this :/
Would using Windows API like `GetMenuItemInfoW` make sense in tests?

compnerd: Yes, I think that validating things using the Win32 APIs is fine.  Many of the tests are already using `@testable import SwiftWin32` just to get access to all sorts of the private things.

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -79,19 +79,53 @@ extension Win32MenuElement {
...
+      } else {
+        imageHandle = nil
+      }
+
+      var submenuHandle: MenuHandle?
+      if let submenu = child as? Menu {
+        submenuHandle = build(submenu)
+      } else {
+        submenuHandle = nil
+      }
```
compnerd, line 116: I think that this can be written more concisely.

```suggestion
      var hSubmenu: MenuHandle? =
          child is Menu ? build(child as! Menu) : nil
```

However, I think that we might be able to generalise this method a bit more and just take a `Menu?` for the parameter?

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -79,19 +79,53 @@ extension Win32MenuElement {
...
+      var wide = child.title.wide
+      var info = wide.withUnsafeMutableBufferPointer {
+        MENUITEMINFOW(cbSize: UINT(MemoryLayout<MENUITEMINFOW>.size),
+                      fMask: UINT(MIIM_FTYPE | MIIM_STATE | MIIM_ID | MIIM_STRING | MIIM_SUBMENU | MIIM_DATA | MIIM_BITMAP),
+                      fType: UINT(MFT_STRING), fState: UINT(MFS_ENABLED), wID: UInt32.random(in: .min ... .max),
+                      hSubMenu: submenuHandle?.value, hbmpChecked: nil, hbmpUnchecked: nil, dwItemData: 0,
+                      dwTypeData: $0.baseAddress, cch: UINT(child.title.count), hbmpItem: imageHandle?.value)
+      }
+      InsertMenuItemW(hMenu.value, UINT(index), true, &info)
+    }
```
compnerd, line 127: I think that some of the existing types can now be removed right?

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -79,19 +79,53 @@ extension Win32MenuElement {
...
+    self.view = view
+    self.menus = []
+  }
+
+  internal func build() -> MenuHandle? {
+    let rootMenu = Menu(children: Array(menus))
+    return build(rootMenu)
+  }
+
+  private func build(_ menu: Menu) -> MenuHandle? {
```
compnerd, line 96: Is there a reason to have the `build(_:Menu)` member?  The vast majority of this logic seems like it should go into `setNeedsRebuild()`: https://compnerd.github.io/swift-win32/SwiftWin32/MenuSystem.html#setneedsrebuild

egorzhdan: Thanks, I reworked this logic with `setNeedsRebuild`.

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -79,19 +79,53 @@ extension Win32MenuElement {
...
-    self.view = view
-    self.menus = []
+
+    for (index, child) in menu.children.enumerated() {
+      let imageHandle: BitmapHandle?
+      if let bitmap = child.image?.bitmap {
+        imageHandle = BitmapHandle(from: bitmap)
+      } else {
+        imageHandle = nil
+      }
```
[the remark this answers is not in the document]
egorzhdan: Hmm, not sure I understand... Why do we need a `==` overload? If the image is `nil` that should work fine as is, since we'll just pass `nil` to the `MENUITEMINFOW` constructor.

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -79,19 +79,53 @@ extension Win32MenuElement {
...
   private var menus: OrderedSet<Menu>
 
   internal init?(for view: View) {
-    self.hMenu = MenuHandle(owning: CreateMenu())
-    if self.hMenu.value == nil {
+    self.view = view
+    self.menus = []
+  }
+
+  internal func build() -> MenuHandle? {
```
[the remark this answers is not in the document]
egorzhdan: I've misunderstood the design idea initially, this overload is now replaced by `setNeedsRebuild`.

egorzhdan:
(I'm going to add the tests soon)

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -22,8 +22,43 @@ internal final class _MenuBuilder: MenuSystem {
...
+      let hSubmenu: MenuHandle?
+      if let submenu = child as? Menu {
+        hSubmenu = MenuHandle(owning: CreateMenu())
+        if hSubmenu?.value == nil {
+          log.warning("CreateMenu: \(Error(win32: GetLastError()))")
+          continue
+        }
+        append(submenu.children, to: hSubmenu!)
+      } else {
+        hSubmenu = nil
```
compnerd, line 39:
```suggestion
```

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -12,8 +12,8 @@ internal final class _MenuBuilder: MenuSystem {
 
   private var menus: OrderedSet<Menu>
 
-  internal init?(for view: View) {
-    self.hMenu = MenuHandle(owning: CreateMenu())
+  internal init?(for view: View, asWindowMenu: Bool = false) {
```
compnerd, line 15: Do we really need the `asWindowMenu` option?  Can you do a context menu on `Window`?  If not, we _could_ get away with `view is WIndow ? CreateMenu() : CreatePopupMenu()`.  I think that tweaking this way might be nicer:

```suggestion
  internal init?(for view: View, popup: Bool = false) {
```

Sources/SwiftWin32/Platform/Win32+Menu.swift
```diff
@@ -22,8 +22,43 @@ internal final class _MenuBuilder: MenuSystem {
...
+
   override func setNeedsRebuild() {
-    // TODO(compnerd) create the actual menus
+    // Remove the old submenus from the menu:
+    for index in (0..<GetMenuItemCount(self.hMenu.value)).reversed() {
+      DeleteMenu(self.hMenu.value, UINT(index), UINT(MF_BYPOSITION))
+    }
+
+    // Create the new submenus and add them to the root menu:
+    append(Array(self.menus), to: self.hMenu)
```
compnerd, line 61: I think that if you do `for (position, item) in self.menus.enumerated() { ... }`, we could avoid the copy for the Array.  I think that just merging the body of `append` into this function is fine, its basically 3 lines more than than this function anyways.

compnerd (commented):
I'm really looking forward to the tests, and then putting this into an example.

Sources/SwiftWin32/Platform/Win32+Menu.swift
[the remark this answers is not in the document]
compnerd: Right you are, just always create the `BitmapHandle`, which you've done.

Sources/SwiftWin32/Platform/Win32+Menu.swift
egorzhdan: We now have a bit more logic in the `if` block, there probably isn't a way to inline it anymore.

Sources/SwiftWin32/Platform/Win32+Menu.swift
egorzhdan: That doesn't compile, do you want me to make `hSubmenu` a `var`?

Sources/SwiftWin32/Platform/Win32+Menu.swift
egorzhdan: Yeap, I think you're right, we don't really need this argument.

Sources/SwiftWin32/Platform/Win32+Menu.swift
egorzhdan: `append` calls itself recursively for submenu construction, so I'm not sure we can trivially merge these functions, but I got rid of the unnecessary copies.

Sources/SwiftWin32/Platform/Win32+Menu.swift
compnerd: Yeah, that sounds good.

compnerd (approved):
I think that this seems good enough to merge, though I think that having a test case for a rebuild clearing out the previous menu would be nice.

PR #627, opened 2021-07-01
View and Controls: remove context menu handling

compnerd:
This was not completed, and more importantly, took over a name that is
meant to be used in subclasses.  This renames the `menu` item to
`hMenu`, and makes it a menu handle.  The construction of the menu needs
to be plumbed through, but this removes the old handling while the new
handling is implemented.

### DIFF
--- a/Sources/SwiftWin32/Platform/Win32+Menu.swift
+++ b/Sources/SwiftWin32/Platform/Win32+Menu.swift
@@ -6,78 +6,6 @@ import WinSDK
 @_implementationOnly
 import OrderedCollections
 
-internal struct Win32Menu {
-  internal let hMenu: MenuHandle
-
-  private let items: [Win32MenuElement]
-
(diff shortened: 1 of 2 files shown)


compnerd:
@egorzhdan given your existing change to flesh out the menu building, I think that this makes sense to clean up while building up the menu builder.  WDYT?  Or am I not thinking through the implementation enough?

egorzhdan:
@compnerd yeah I think the `Win32Menu` mechanism can be removed at this point. The `_MenuBuilder` will need to be tweaked a bit to support both context menus & window menus (probably just a matter of calling `CreateMenu` vs `CreatePopupMenu`) but that can be done in the scope of my PR.

compnerd:
Great!  Yeah, I think that an internal flag to convert the menu makes sense and unifies the behaviour.

PR #644, opened 2021-07-12
View Controllers: extend, document, open

compnerd:
Add additional documentation for `ViewController` to capture many of the
default value reasonining.  Extend the interface to add many of the
missing interfaces.  Finally, mark many of the interfaces as open.  It
is unclear if the split view controller integration is meant to be open.


compnerd:
CC: @egorzhdan @tomaszpieczykolan

This defines the rest of the `ViewController` interface and open's it up for other modules to re-open the type.  However, there are many missing interfaces that are needed to actually get this to build.  Getting the test coverage from @tomaszpieczykolan's other change would also be nice, but can be split off.

PR #656, opened 2021-08-07
View Controllers: add `SplitViewController`

compnerd:


A3ew3 (changes requested):
Sources/SwiftWin32/View Controllers/SplitViewController.swift``

Issue #768, opened 2023-10-27
Can not build with Swift 5.9.1


tonka3000:
I tried to build the project with the 5.9.1 build from https://www.swift.org/download but I always get the same error with CoreAnimation.

Any idea what is wrong here? Static vs dynamic build?

```
[main] Building folder: swift-win32 
[build] Starting build
[proc] Executing command: C:\Users\tbam\Downloads\cmake-3.27.7-windows-x86_64\bin\cmake.EXE --build C:/sandbox/swift/swift-win32/build/x86_64-unknown-windows-msvc/Release --target all
[build] [1/6] Linking Swift static library lib\CoreAnimation.lib
[build] FAILED: lib/CoreAnimation.lib Sources/SwiftWin32/CMakeFiles/CoreAnimation.dir/CoreAnimation/Transform3D.swift.obj swift/CoreAnimation.swiftmodule 
[build] cmd.exe /C "cd . && C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe -j 6 -num-threads 6 -emit-library -static -o lib\CoreAnimation.lib -module-name CoreAnimation -module-link-name CoreAnimation -emit-module -emit-module-path swift\CoreAnimation.swiftmodule -emit-dependencies  -sdk C:/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk -O -wmo -libc MD -output-file-map Sources\SwiftWin32\CMakeFiles\CoreAnimation.dir\Release\output-file-map.json  C:\sandbox\swift\swift-win32\Sources\SwiftWin32\CoreAnimation\Transform3D.swift    && cd ."
[build] error: link command failed with exit code 1 (use -v to see invocation)
[build] /usr/bin/link: extra operand '/OUT:lib\\CoreAnimation.lib'
[build] Try '/usr/bin/link --help' for more information.
[build] error: fatalError
[build] [2/6] Linking Swift shared library bin\Collections.dll
[build] ninja: build stopped: subcommand failed.
[proc] The command: C:\Users\tbam\Downloads\cmake-3.27.7-windows-x86_64\bin\cmake.EXE --build C:/sandbox/swift/swift-win32/build/x86_64-unknown-windows-msvc/Release --target all exited with code: 1
[driver] Build completed: 00:00:00.574
[build] Build finished with exit code 1
```

compnerd:
[...]

PR #769, opened 2023-10-28
Update cmake section in readme with required env-vars

tonka3000:
Add more information about compiling with cmake as mentioned in [#768 ](https://github.com/compnerd/swift-win32/issues/768#issuecomment-1783268566)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,11 @@ Following environment should be set/adjusted before run the above commands (the 
 
   Command Prompt: `set SDKROOT=C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk;%PATH%`
 
-- `PATH`
+- `Path`
 
-  The `PATH` environment variables needs to be extended that CMake can find the swift compiler. This is only required when this is not already the case. When you type `where swiftc` and see no path to swiftc the following command is required.
+  The `Path` environment variables needs to be extended that CMake can find the swift compiler. This is only required when this is not already the case. When you type `where swiftc` and see no path to swiftc the following command is required.
 
-  Command Prompt: `set PATH=C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin;%PATH%`
+  Command Prompt: `set Path=C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin;%Path%`
 
 
 
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ Following environment should be set/adjusted before run the above commands (the 
 - Visual Studio compiler toolchain
 
   The CMake commands use the `mt` tool which requires a Visual Studio compiler environment (Build Tools should be fine too).
-  e.g. if you have Visual Studio 2019 installed you can create a command prompt by click on `x64 Native Tools Command Prompt for VS2019` from the Visual Studio 2019 installation in the Windows Start Menu or call `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat` in the command prompt.
+  e.g. if you have Visual Studio 2019 installed you can create a command prompt by click on `x64 Native Tools Command Prompt for VS2019` from the Visual Studio 2019 installation in the Windows Start Menu.
   
-  Higher versions of Visual Studio or the corresponding Build Tools should be fine too.
+  Newer versions of Visual Studio or the corresponding Build Tools should be fine too.
 
   Another way of activating a Visual Studio environment is the tool [vswhere](https://github.com/microsoft/vswhere).
 
@@ -45,13 +45,13 @@ Following environment should be set/adjusted before run the above commands (the 
 
   \>= 5.10 `SDKROOT` needs to be `%LocalAppData%\Programs\Swift\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
 
-  < 5.10 `SDKROOT` needs to be `C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
+  < 5.10 `SDKROOT` needs to be `%SystemDrive%\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
 
   Examples for < 5.10:
 
-  Command Prompt: `set SDKROOT=C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
+  Command Prompt: `set SDKROOT=%SystemDrive%\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
 
-  PowerShell: `$env:SDKROOT="C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk"`
+  PowerShell: `$env:SDKROOT="%SystemDrive%\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk"`
 
 - `Path`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,25 @@ ninja -C build SwiftWin32 UICatalog
 %CD%\build\bin\UICatalog.exe
 ```
 
+Following environment variables should be set/adjusted before run the above commands (the shown paths assume that you use compiler from the official swift installers from [swift.org](https://www.swift.org/download/)).
+
+- `SDKROOT`
+
+  This should be the path to the Windows Platform SDK from the Swift library.
+
+  Command Prompt: `set SDKROOT=C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk;%PATH%`
```
compnerd, line 37: This should be set automatically by the installer. If it is not, we should investigate what is going on. This is only needed during the build, not the run.

README.md
```diff
@@ -28,6 +28,25 @@ ninja -C build SwiftWin32 UICatalog
...
+
+- `PATH`
```
compnerd, line 39: Windows spells this `Path` but it is case insensitive so it ends up being the same.

README.md
```diff
@@ -28,6 +28,25 @@ ninja -C build SwiftWin32 UICatalog
...
+
+  The `PATH` environment variables needs to be extended that cmake can find the swift compiler. This is only required when this is not already the case. When you type `where swiftc` and see no path to swiftc the following command is required.
+
+  Command Prompt: `set PATH=C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin;%PATH%`
```
compnerd, line 43: The tools path should also be setup by the installer. If this is not the case, we should investigate what is going on.

README.md
```diff
@@ -28,6 +28,25 @@ ninja -C build SwiftWin32 UICatalog
...
+
+- Visual Studio compiler toolchain
+
+  The cmake commands use the `mt` tool which require a fully bootstrapped compiler environment. You can create a command prompt by click on `x64 Native Tools Command Prompt for VS2019` from the Visual Studio 2019 installation in the Windows Start Menu or call `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat` in the command prompt.
```
compnerd, line 47:
```suggestion
The CMake commands use the `mt` tool which requires a Visual Studio compiler environment. You can create a command prompt by click on `x64 Native Tools Command Prompt for VS2019` from the Visual Studio 2019 installation in the Windows Start Menu or call `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat` in the command prompt.
```

compnerd: I wonder if we should instruct people to use `vswhere` to find the latest environment. This would be better than assuming that VS2019 is installed (VS2022 could be installed), and more importantly, for the docker environment, the expectation is that we will have VS Build Tools, not VS2019.

README.md
tonka3000: After I restarted my machine it was the case. I assume that is the classic "Open from Windows Explorer" after the an installation, so the explorer don't got the new env-vars.

I would change it to "if it is not set" or should I remove it completely?

btw: Naming the system wide variable only `SDKROOT` sound like a high potential for a naming collision. `SWIFT_SDKROOT` would be a better choice in my opinion. I had a lot of tool which make similar decision and I always need to find workarounds for this type of issue.

README.md
tonka3000: Correct. This is something which confuse me for decades 😄

README.md
tonka3000: Same as on the other part, after restart (a restart of the explorer process would do the same) it was set.

README.md
tonka3000: Mention `vswhere` could help, but it depends on the skill set of the user as well. The `x64 Native Tools Command Prompt for VS20xx` is super easy to start play around. `vswhere` could be a little more complicated but is the better automate able solution.

I added a small sentence about `vswhere`. Let me know if I should add more (I personally don't use `vswhere` directly, I use conan or cmake to bootstrap my environments).

README.md
tonka3000: I've added the sentence that this should be the case if the installer was used.

README.md
tonka3000: I've added the sentence that this should be the case if the installer was used.

README.md
```diff
@@ -28,6 +28,31 @@ ninja -C build SwiftWin32 UICatalog
...
+
+- `SDKROOT`
+
+  This should be the path to the Windows Platform SDK from the Swift library. This should be already set by the official swift installer. If not set the env var.
+
+  Command Prompt: `set SDKROOT=C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk;%PATH%`
+
+- `Path`
+
+  The `Path` environment variables needs to be extended that CMake can find the swift compiler. This is only required when this is not already the case. When you type `where swiftc` and see no path to swiftc the following command is required.
```
compnerd, line 50:
```suggestion
  The `Path` environment variables needs to be extended to allow the Swift compiler to be found. This is only required when this is not already the case. When you type `where swiftc` and see no path to `swiftc` the following command is required.
```

compnerd: `where swiftc` doesn't really work with PowerShell, so we would need to be more explicit. I think that perhaps we should just add a troubleshooting guide instead? We could put that on the Wiki so that we can make changes to this easier and grow it over time.

README.md
compnerd: Yeah, I think that is it - that starting from explorer would not get the new environment variables. I always just launch a new terminal instance, so I never noticed this and never thought to include that in the documentation.

Yeah, I agree with you about the naming. I would honestly prefer a nicer approach than using an environment variable. Note that the environment variable is _optional_. If you do not have it set, you need to explicitly pass the path along to the compiler for the `-sdk` parameter (which is implicitly set if `SDKROOT` is set).  However, the use of `SDKROOT` does mean that we have similar behaviour across platforms.

One final issue - the path is not guaranteed. It uses `%SystemDrive%` until 5.10.  With 5.10 the installer is now per-user and will be installed to `%LocalAppData%\Programs\Swift`, with the SDKs under `%LocalAppData%\Programs\Swift\Platforms`, toolchains under `%LocalAppData%\Programs\Swift\Toolchains`, and the runtime under `%LocalAppData%\Programs\Swift\Runtimes`.

tonka3000: I added powershell commands as well as 5.10 variables

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
 %CD%\build\bin\UICatalog.exe
 ```
 
+Following environment should be set/adjusted before run the above commands (the shown paths assume that you use compiler from the official swift installers from [swift.org](https://www.swift.org/download/)).
```
compnerd, line 31:
```suggestion
Following environment variables should be set/adjusted before run the above commands (the shown paths assume that you use compiler from the official swift installers from [swift.org](https://www.swift.org/download/)).
```

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+
+- Visual Studio compiler toolchain
+
+  The CMake commands use the `mt` tool which requires a Visual Studio compiler environment (Build Tools should be fine too).
```
compnerd, line 35: SPM also needs to use it, it just has to be done manually.

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+  e.g. if you have Visual Studio 2019 installed you can create a command prompt by click on `x64 Native Tools Command Prompt for VS2019` from the Visual Studio 2019 installation in the Windows Start Menu or call `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Auxiliary\Build\vcvars64.bat` in the command prompt.
```
compnerd, line 36: Why Professional? I may be using Enterprise or Community. This is why I think that we should really just provide a VSWhere invocation instead.

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+  
+  Higher versions of Visual Studio or the corresponding Build Tools should be fine too.
```
compnerd, line 38: Higher is confusing here - because the enterprise SKU is considered higher. Perhaps "Newer"?

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+
+  Another way of activating a Visual Studio environment is the tool [vswhere](https://github.com/microsoft/vswhere).
+
+- `SDKROOT`
+
+  This should be the path to the Windows Platform SDK from the Swift library. This should be already set by the official swift installer. If not set the env var.
+
+  \>= 5.10 `SDKROOT` needs to be `%LocalAppData%\Programs\Swift\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
+
+  < 5.10 `SDKROOT` needs to be `C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
```
compnerd, line 48: Can we use `%SystemDrive%` instead of `C:` as the installer does?

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+
+  Examples for < 5.10:
+
+  Command Prompt: `set SDKROOT=C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk`
+
+  PowerShell: `$env:SDKROOT="C:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk"`
```
compnerd, line 54: Similar

README.md
```diff
@@ -28,6 +28,49 @@ ninja -C build SwiftWin32 UICatalog
...
+
+- `Path`
+
+  The `Path` environment variables needs to be extended to allow the Swift compiler to be found.
+  This is only required when this is not already the case.
+  When you type `where swiftc` (`(Get-Command swiftc).Path` for PowerShell) and see no path to `swiftc` the following command is required.
+
+  \>= 5.10 the `Path` needs to extended with `%LocalAppData%\Programs\Swift\Runtimes\0.0.0\usr\bin;%LocalAppData%\Programs\Swift\Toolchains\0.0.0+Asserts\usr\bin`
```
compnerd, line 62: The 0.0.0 will be actual release number of actual releases and 0.0.0 for snapshots or developer builds.

tonka3000: I've added `5.9.0` as example

README.md
tonka3000: I removed the path to vsvars64.bat. So now it is the same as the section from SPM.
I agree that vswhere would be better but I never used it.

README.md
tonka3000: Not quite sure what you mean here?
There is already the mention of the `mt` in the SPM section.
Should I rephrase here?

compnerd (approved):

PR #782, opened 2024-04-05
Remove symlinks

### LINKED ISSUE #781: Building fails with an internal error: symlinks not supported
When trying to build or resolve package, the process exits with this error:
```shell
E:\swift\happyworkwin> swift package resolve
Updating https://github.com/compnerd/swift-win32.git
Updated https://github.com/compnerd/swift-win32.git (0.78s)
error: the package manifest at '\Package@swift-5.8.swift' cannot be accessed (InternalError(description: "Internal error. Please file a bug at https://github.com/apple/swift-package-manager/issues with this info. symlinks not supported")) in https://github.com/compnerd/swift-win32.git
E:\swift\happyworkwin>
```
As the error message says, this is probably due to `Package@swift-5.8.swift` just being a symlink to `Package@swift-5.7.swift`. This seems like it's not needed, as I use Swift 5.9.2 which automatically failed back to the (broken) 5.8 `Package.swift`.

libewa:
Removed `Package.swift` symlinks as they break SwiftPM and are not needed, as it already fails back to the latest version supported by the local Swift package tools.
Closes #781

### DIFF
--- a/Package@swift-5.5.swift
+++ b/Package@swift-5.5.swift
@@ -1,1 +1,0 @@
-Package@swift-5.4.swift
--- a/Package@swift-5.6.swift
+++ b/Package@swift-5.6.swift
@@ -1,1 +1,0 @@
-Package@swift-5.4.swift
--- a/Package@swift-5.8.swift
+++ b/Package@swift-5.8.swift
@@ -1,1 +1,0 @@
-Package@swift-5.7.swift


compnerd (changes requested):
This will break CI as we verify that the package remains compatible with different versions. These files must be retained, and I would rather not duplicate the content and have to remember to edit each version.

